PR #2, opened 2026-01-09
FLPATH-2960: Add OpenAPI spec and code generation [1/5]

gciavarrini:
Adds AEP-compliant OpenAPI spec and oapi-codegen generated code.

## Changes
- OpenAPI spec with RFC 7807 error responses
- Generated types, server, and client code
- Makefile targets for code generation
- CI workflow to check generated files are in sync

## PRs plan
1. **This PR** - OpenAPI spec and code generation
2. Core infrastructure (config, server setup)
3. Data layer (store, models)
4. Business logic (service, handlers)
5. Registration client library

## Fixes
Partially addresses [FLPATH-2960](https://issues.redhat.com/browse/FLPATH-2960)


api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+              schema:
+                $ref: '#/components/schemas/Error'
+        default:
+          description: Unexpected error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+  /providers/{providerId}:
```
jenniferubah, line 148: Should this be `/providers/providerName`? If the provider performs self-registration, then I'm not sure how it would persists the generated DCM uuid ( `providerId`). With the name, it already knows and no need for the providers to keep another record of the uuid. What do you think?

ygalblum: In order to adhere to the AEP spec, it must be the UUID

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+  - url: /api/v1
+
+tags:
+  - name: provider
+    description: Service Provider management operations
+  - name: health
+    description: Health check operations
+
+paths:
+  /health:
```
ygalblum, line 24: In order to adhere to the AEP spec, the health request must return a defines resource.
The API should be something like:
```yaml
/health:
  get:
    summary: Health check
    operationId: GetHealth
    description: Health check for DCM Placement API
    responses:
      '200':
        description: OK
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/health'
```
The resource should be something like:
```yaml
health:
  type: object
  x-aep-resource:
    singular: health
    plural: healths
    singleton: true
    type: placement.dcm.io/health
    patterns:
    - health
  properties:
    status:
      type: string
      description: Health status
      example: "healthy"
    path:
      type: string
      description: Canonical path of the resource
      example: "health"
      readOnly: true
```

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+    get:
+      tags:
+        - provider
+      summary: List all providers
+      operationId: ListProviders
+      description: Returns a list of registered service providers with optional filtering
+      parameters:
+        - name: type
+          in: query
+          required: false
```
ygalblum: What's the default value of `required`? If `false`, then you don't need it here. If `true`, then you need it in the other parameters (since they are not required either)

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+          description: Filter providers by service type
+          schema:
+            type: string
+        - name: max_page_size
```
ygalblum, line 52: You should define `min` (1), `max` (whatever you think) and a `default` (I think equal to `max`). This way there is never a request that is expected to return an enormous number of resources (I know it's the SP list, which is not supposed to be that big, but still for best practices)

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+              schema:
+                $ref: '#/components/schemas/Error'
+        default:
+          description: Unexpected error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+    post:
```
ygalblum, line 85: According to the AEP spec, there should be a single component for all operations on the same endpoint. So, there should be a single component `Provider` for all operations (`post`, `get` and `put`).

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+      operationId: CreateProvider
+      description: |
+        Register a new service provider or update an existing one.
+        Registration is idempotent:
+        - If name does not exist, a new SP entry is created
+        - If name exists with same/no providerID, the entry is updated
+        - If name exists with different providerID, registration fails (conflict)
+        - If providerID exists with different name, registration fails (conflict)
+      parameters:
+        - name: id
```
ygalblum, line 98: Once using a single component type, there's no need for this parameter

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+    put:
+      tags:
+        - provider
+      summary: Update a Service Provider
+      operationId: ApplyProvider
+      description: |
+        Update an existing service provider.
+        The providerID in the path must match an existing provider.
+      parameters:
+        - name: providerId
```
ygalblum, line 198: The `id` is already part of the `Provider` in the request body

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ProviderResponse'
+        '201':
+          description: Provider created (new registration)
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ProviderResponse'
```
ygalblum: According to the AEP spec, the return payload of a `post` request should be the updated component, not a subset of it

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+        default:
+          description: Unexpected error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+components:
+  schemas:
+    ProviderRegistration:
```
ygalblum: As stated, this should be merged with the `Provider` component

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+        endpoint:
+          type: string
+          format: uri
+          description: Full endpoint URL where the provider API is accessible
+          example: "https://sp1.example.com/api/v1/vm"
+        service_type:
+          type: string
+          description: Type of service this provider offers
+          example: "vm"
+          enum:
```
ygalblum: The values cannot be predefined at the API level because new types can be added dynamically

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+        create_time:
+          type: string
+          format: date-time
+          description: Timestamp when the provider was first registered
+        update_time:
+          type: string
+          format: date-time
+          description: Timestamp when the provider was last updated
+
+    ProviderResponse:
```
ygalblum: As commented, this should be replaced by the `Provider`

api/v1alpha1/openapi.yaml
gciavarrini: @ygalblum is spot on here. That's why i'm using the `providerId`

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+        total_storage:
+          type: string
+          description: Total storage available
+          example: "2TB"
+        total_node:
+          type: integer
+          description: Total number of nodes
+          example: 100
+
+    Provider:
```
ygalblum, line 322: While I'm not sure it's currently being validated, the AEP spec mandates that components names start with a lower case: https://aep.dev/4/#type-name

ygalblum (commented):
The API itself and the code generation seem fine by me. However, as we would like to adhere to the AEP spec (as much as we can), all my comments are targeted to it.

In addition, to make sure we pass its validation, we should add a makefile target for it and make sure it passed in the CI before merging.
See: https://github.com/dcm-project/dcm-placement-api/blob/52faac8d314fe1780f5c1166745589660b7ba228/Makefile#L13 for the Makefile and https://github.com/dcm-project/dcm-placement-api/blob/52faac8d314fe1780f5c1166745589660b7ba228/.spectral.yaml for the configuration.

api/v1alpha1/openapi.yaml
jenniferubah: hmm , am i missing something? because the linter passes for me (see [here](https://github.com/dcm-project/kubevirt-service-provider/pull/10/files)) when I removed the `format: uuid `and changed the path param to `vmName` and it works. Also in the [aep doc](https://aep.dev/122/), I couldn't find it.

.spectral.yaml
```diff
@@ -0,0 +1,14 @@
+extends:
+  - https://raw.githubusercontent.com/aep-dev/aep-openapi-linter/main/spectral.yaml
+
+# /health is an infrastructure endpoint, not a business resource.
```
ygalblum: Makes sense

api/v1alpha1/openapi.yaml
ygalblum: I saw that you've excluded `health` from the check. I think that's good enough

api/v1alpha1/openapi.yaml
ygalblum: I think I might be wrong here. I've resolving this comment for now

api/v1alpha1/openapi.yaml
ygalblum: AEP does not define the name of the field and I guess the linter does not mandate it either. But, in the sepc's essence, each resource has an `id` (which can be passed by the client or created by the server) during creation. This `id` is then used in the other methods.

Having said that, I'm not sure that anything is stopping us from not using an `id` and actually using `name` as the `id`. If you look here: https://aep.dev/122 the resource IDs can be a name.

So, I'm not sure that we can't just use the `name` here

ygalblum: The advantage of using the name is like @jenniferubah wrote, the provider will not need to store an id. The disadvantage I see is that if two SPs register with the same name, the second one will override the first one (since we defined the API as idempotent. See: https://aep.dev/155/

gciavarrini: @jenniferubah You're right aep-122 allows both approaches. 
Using name in path is simpler but using UUID gives immutability if an SP renames itself.
This was discussed during the enhancement review, and the decision was to use UUID for immutability so that other DCM components referencing a provider won't break on rename. 
I'm open to revisiting this, if you prefer using name but we need to update the enhancement first. 

What do you think? @ygalblum @jenniferubah

api/v1alpha1/openapi.yaml
gciavarrini: Makes sense. Added minimum: 1, maximum: 100, and default: 100.

api/v1alpha1/openapi.yaml
gciavarrini: Good catch! OpenAPI 3.0 defaults query params to `required: false`, this redundant declaration is a leftover :)

api/v1alpha1/openapi.yaml
gciavarrini: Done. POST now returns the full Provider resource.

api/v1alpha1/openapi.yaml
gciavarrini: Addressed using a single Provider schema

api/v1alpha1/openapi.yaml
gciavarrini: See previous comments, removed `ProviderResponse` use `Provider`

api/v1alpha1/openapi.yaml
gciavarrini: Looking at AEP-4, the lowercase requirement applies to the resource type string (e.g `serviceprovider.dcm.io/provider`), not OpenAPI schema names. Our `x-aep-resource` type fields are already lowercase. The linter passes without warnings on schema names. . OpenAPI typically uses PascalCase for schemas, I think we can keep it as it is. Makes sense?

.spectral.yaml
gciavarrini: edited as you suggested in https://github.com/dcm-project/service-provider-api/pull/2#discussion_r2676841340 let me know what you prefer. I don't have a strong opinion

api/v1alpha1/openapi.yaml
gciavarrini: i'm using the `Provider` schema, however, AEP-133 says if user-specified IDs are supported, the id must be a query parameter. The linter enforces this. So I'm keeping the query param for AEP compliance. is it ok? Am i misunderstanding somenthing?

api/v1alpha1/openapi.yaml
jenniferubah: Let's leave it as is for now. We can revist it later in the future if needed

api/v1alpha1/openapi.yaml
ygalblum: OK. Kinda annoying that AEP uses a different case (kebab-case). But, yes, makes sense

api/v1alpha1/openapi.yaml
ygalblum: No, you're right. I'm the one who misunderstood it. Sorry

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,442 @@
...
+  - name: health
+    description: Health check operations
+
+paths:
+  /health:
+    get:
+      tags:
+        - health
+      summary: Health check
+      operationId: GetHealth
```
gabriel-farache: I know that @jenniferubah asked about it in others PRs for other services: about the case, should be use lowerCamelCase or PascalCase? Then we should be consistent across all services spec

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,442 @@
...
+        - name
+        - endpoint
+        - service_type
+      properties:
+        id:
+          type: string
+          format: uuid
+          description: Unique identifier for the Service Provider
+          example: "123e4567-e89b-12d3-a456-426614174000"
+        path:
```
jenniferubah, line 342: Just for my understanding, what can the `path` field be used for and can it be provided (optionally) during registration?

jenniferubah (commented):
Overal, looks good. We mainly need to decide what naming convention to adopt and then I'm good.

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,442 @@
...
+        default:
+          description: Unexpected error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+components:
+  schemas:
+    ProviderMetadata:
```
ygalblum, line 281: Should we add `additionalProperties: true` to allow for untyped information?

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,442 @@
...
+          example: "2TB"
+        total_node:
+          type: integer
+          description: Total number of nodes
+          example: 100
+
+    Provider:
+      type: object
+      description: Full provider resource representation
+      x-aep-resource: true
```
ygalblum: Should we define the entire structure for `x-aep-resource` instead of just setting it to `true` according to https://aep.dev/4/#annotating-resource-types?

api/v1alpha1/openapi.yaml
ygalblum: It's an AEP requirement. It's supposed to be calculated by the server and marked as `readOnly`

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,442 @@
...
+      required:
+        - name
+        - endpoint
+        - service_type
+      properties:
```
ygalblum, line 335: Should we define the `readOnly` items as such (`id`, `path`, `create_time`, `update_time`, did I miss something)?

api/v1alpha1/openapi.yaml
gabriel-farache: you mean other properties that are not defined below?

api/v1alpha1/openapi.yaml
gciavarrini: Agree, we should align across all DCM repos.
I don't have a strong opinion and openapi doesn't have a direction for this.

In the [petstore](https://petstore3.swagger.io/api/v3/openapi.json) which is the official OpenAPI example, they use `lowerCamelCase`.
 I'd suggest using `lowerCamerCase` (so renaming `GetHealth` to `getHealth`).
 
 WDYT? 
 @jenniferubah

api/v1alpha1/openapi.yaml
gciavarrini: yes, it [AEP-122](https://aep.dev/122/). 
Added `readOnly`.

api/v1alpha1/openapi.yaml
gciavarrini: Updated to match AEP-4.
Also added a local spectral rule to catch `x-aep-resource: true`: the linter skips validation when set to `true`. Updated `make check-aep` to fail on warnings too.

api/v1alpha1/openapi.yaml
gciavarrini: Correct.
Added `readOnly: true` to `path`, `status`, `create_time`, and `update_time`.
While `id` is not readonly since service providers can provide it via query parameter per AEP-133 for idempotency.

api/v1alpha1/openapi.yaml
gciavarrini: resolved in https://github.com/dcm-project/service-provider-manager/pull/2#discussion_r2681493438

api/v1alpha1/openapi.yaml
gciavarrini: Good suggestion. 
Added `additionalProperties: true` to allow providers to include custom metadata fields beyond the defined ones.

api/v1alpha1/openapi.yaml
jenniferubah: Thanks!

api/v1alpha1/openapi.yaml
jenniferubah: Fine by me 👍🏾

jenniferubah (approved):

api/v1alpha1/openapi.yaml
ygalblum: IIUC the reason for having the `id` in the `POST`'s query parameter is to allow the `id` in the payload to be `readOnly`. Otherwise, the user could have just passed it as part of the payload. No?

api/v1alpha1/openapi.yaml
ygalblum: Great. @gabriel-farache I'm leaving this comment for you to either respond or resolve

gabriel-farache (approved):

api/v1alpha1/openapi.yaml
gciavarrini: True, thanks for catching it. 
Since `id` is passed via query param per AEP-133, marking it `readOnly` in the schema avoids conflicts between query param and body values. Added `readOnly: true` to `id`.
Also updated the enhancement to reflect this: dcm-project/enhancements#18

ygalblum (approved):
Not sure what the `.gitattributes` does. Anyhow LGTM

gciavarrini:
> Not sure what the `.gitattributes` does. Anyhow LGTM

It's related to this discussion https://github.com/dcm-project/catalog-manager/pull/1#discussion_r2689390442
It simply collapses generated files in GH UI, so the reviewers are not distracted/confused by auto generated files.

PR #3, opened 2026-01-09
FLPATH-3006: Add core infrastructure [2/5]

gciavarrini:
# Add core infrastructure with configuration and HTTP server

Implements foundational infrastructure for DCM Service Provider Manager including environment-based configuration, HTTP server with graceful shutdown, and production-ready operational patterns.

# Add core infrastructure with configuration and HTTP server

Implements foundational infrastructure for DCM Service Provider Manager including environment-based configuration, HTTP server with graceful shutdown, and production-ready operational patterns.

## PR Series
1. Project scaffolding
2. **This PR** - Core infrastructure  
3. OpenAPI spec + code generation
4. Client library
5. Business logic + handlers

## Fixes
https://issues.redhat.com/browse/FLPATH-3006

PR #4, opened 2026-01-09
FLPATH-3007: Add data layer for SP registrations [3/5]

gciavarrini:
Adds persistence layer for Service Provider registrations.

## Changes
- Provider database model with GORM tags
- Store interface and provider repository (CRUD operations)
- Database initialization (PostgreSQL/SQLite support)
- Tests with Ginkgo/Gomega

## PR Series
1. OpenAPI spec + code generation
2. Core infrastructure
3. **This PR** - Data layer
4. Business logic + handlers
5. Client library

## Fixes
https://issues.redhat.com/browse/FLPATH-3007

---
Depends on PR #3


internal/store/db.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/dcm-project/service-provider-api/internal/config"
+	"github.com/dcm-project/service-provider-api/internal/store/model"
+	"gorm.io/driver/postgres"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+)
+
+func InitDB(cfg *config.Config) (*gorm.DB, error) {
```
machacekondra, line 16: Do we want to use shared database for all(in this case I think placement+SP) services or database per service?

internal/store/model/provider.go
```diff
@@ -0,0 +1,30 @@
...
+	ServiceType  string         `gorm:"column:service_type;not null"`
+	Endpoint     string         `gorm:"column:endpoint;not null"`
+	Path         string         `gorm:"column:path"`
+	RegionCode   string         `gorm:"column:region_code"`
+	Zone         string         `gorm:"column:zone"`
+	Status       string         `gorm:"column:status"`
+	TotalCPU     int            `gorm:"column:total_cpu"`
+	TotalMemory  string         `gorm:"column:total_memory"`
+	TotalStorage string         `gorm:"column:total_storage"`
+	TotalNode    int            `gorm:"column:total_node"`
```
machacekondra: I just wonder if we want to add those parameters now, or if we wanna add them at a point we implement the thing, that report those fields. Same for other fields.

internal/store/provider.go
```diff
@@ -0,0 +1,109 @@
...
+	if err := s.db.WithContext(ctx).First(&provider, id).Error; err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return nil, ErrProviderNotFound
+		}
+		return nil, err
+	}
+	return &provider, nil
+}
+
+func (s *ProviderStore) GetByName(ctx context.Context, name string) (*model.Provider, error) {
```
machacekondra, line 103: Do we want rather a method like `Get(ctx, filter Filter)` ?

internal/store/db.go
```diff
@@ -0,0 +1,64 @@
...
+
+	sqlDB, err := db.DB()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get underlying db: %w", err)
+	}
+	sqlDB.SetMaxIdleConns(10)
+	sqlDB.SetMaxOpenConns(100)
+
+	// Auto-migrate schema
+	if err := db.AutoMigrate(&model.Provider{}); err != nil {
```
machacekondra, line 58: We need to agree on tooling we will do the sql migration.

cmd/service-provider-api/main.go
```diff
@@ -1,5 +1,93 @@
...
+}
+
+func (notImplementedResponse) VisitGetProviderResponse(w http.ResponseWriter) error {
+	w.WriteHeader(http.StatusNotImplemented)
+	return nil
+}
+
+func (notImplementedResponse) VisitApplyProviderResponse(w http.ResponseWriter) error {
+	w.WriteHeader(http.StatusNotImplemented)
+	return nil
```
machacekondra, line 95: Hmm, aren't those part of `server.gen.go`? Why do we need them here?

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+    put:
+      tags:
+        - provider
+      summary: Update a Service Provider
+      operationId: ApplyProvider
```
machacekondra: Just wonder about the name `Apply` is there any reason behind it?

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+        path:
+          type: string
+          description: Resource path identifier
+          example: "providers/123e4567-e89b-12d3-a456-426614174000"
+        status:
+          type: string
+          description: Registration status
+          enum:
+            - registered
+            - updated
```
machacekondra: Updated means what exactly?

internal/store/db.go
gciavarrini: Good question.
I assumed database per service for now so that each service manages its own schema and migrations independently. I selected the easiest approach :) 
If we want a shared database, we'd need to coordinate migrations and schema changes across services.
Do you think we can use this approach for v1 and then revisit the decision? Or do you have a preference?

internal/store/model/provider.go
gciavarrini: These fields come from the enhancement doc ([sp-registration-flow.md](https://github.com/dcm-project/enhancements/blob/c605a943cc943792cb01734fc3908341716b4538/enhancements/sp-registration-flow/sp-registration-flow.md?plain=1#L153-L160)). 
I included them now to match the API spec, but we can defer if you prefer. 
Should I remove those fields so that we can and add them when we'll implement the reporting?

machacekondra: I don't mind, whatever you prefer, but it would be a dead code, until it's implemented. I guess I just like more when it's added when it's needed, but I fine with whatever approach we do.

internal/store/db.go
machacekondra: Well in our case, I guess the best would be DB per service. But I am fine to use shared one, if anyone has reasonable objection to use it.

internal/store/provider.go
gciavarrini: You're right and since registration is per `service type`, we need `name+service_type`, not just name.
I'll add a simple filter struct with optional fields for exact matches wihout operators (e.g, `gte`, `lte`, ...).
So we'll have `Get(ctx, filter{Name: &name, ServiceType: &serviceType})`.
In this way, we keep things simple while supporting filtering. 
Is it reasonable?

internal/store/db.go
gciavarrini: Agree.
GORM `AutoMigrate` is temporary for development. For prod we should use proper migrations.
Should we decide on the tooling now, or keep `AutoMigrate` for this PR and add proper migrations in a follow-up?

api/v1alpha1/openapi.yaml
gciavarrini: AEP requires it. I don't like it as well but... 🤷‍♀️

See[AEP-137](https://aep.dev/tooling/openapi-linter/rules/0137/#aep-137-operation-id)

cmd/service-provider-api/main.go
gciavarrini: This is a temporary solution.
The Visit methods in `server.gen.go` are for response types defined in the OpenAPI spec (`200`, `201`, `400`, etc.). Since `501` isn't in the spec, no `501` response type was generated. I created `notImplementedResponse` as a temporary stub to satisfy `StrictServerInterface` until we wire up real handlers in the next PR.
Once handlers are implemented, this will be removed. Is it ok to keep it here for now?

api/v1alpha1/openapi.yaml
gciavarrini: The idea is to track that  the provider was re-registered (idempotent update) rather than newly created.
As per [enhancement](https://github.com/dcm-project/enhancements/blob/c605a943cc943792cb01734fc3908341716b4538/enhancements/sp-registration-flow/sp-registration-flow.md?plain=1#L267) doc, when an SP re-registers with the same name, it updates the existing entry and returns status `updated` while `registered` means a new entry was created. wdyt?

internal/store/model/provider.go
jenniferubah: I prefer the approach of adding code when needed. Just my preference. Less to review as well.

internal/store/db.go
machacekondra: For this PR it's fine. Just made a comment, so we don't forget to have a discussion about it.

cmd/service-provider-api/main.go
machacekondra: Ah, ok, sure!

api/v1alpha1/openapi.yaml
machacekondra: Hmm, I just wonder what's the value for the user to see this? If I update it and then update it to original values, it's not really changed, but says updated.

Depends what those values are used for (those that can be updated). Who is interested in knowing if it was updated? Is it policy engine or user?

api/v1alpha1/openapi.yaml
```diff
@@ -0,0 +1,506 @@
...
+    This API implements the Service Provider registration flow as defined in the
+    DCM enhancement proposal.
+  title: Service Provider API
+  version: v1alpha1
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+
+servers:
+  - url: /api/v1
```
machacekondra: @gciavarrini FYI: @ygalblum is using `v1alpha1`, please make sure you agree on the version

internal/store/db.go
```diff
@@ -0,0 +1,64 @@
...
+	"gorm.io/driver/postgres"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+)
+
+func InitDB(cfg *config.Config) (*gorm.DB, error) {
+	var dialector gorm.Dialector
+
+	if cfg.Database.Type == "pgsql" {
```
ygalblum, line 19: I should have probably mentioned this in the previous PR. But, should something (e.g. the `config` module) validate the value provided by the user? I mean any value that's not `pgsql` ends up with `sqlite`

internal/store/model/provider.go
ygalblum: But, doesn't this comment apply to the entire model? I mean, this PR does not access or set any of these fields.

internal/store/db.go
ygalblum: I like the per service DB approach. I think it does impact that interaction between the user, the `CatalogManager` and `PlacementManager`. We'll need to discuss this

internal/store/provider.go
ygalblum: Will you still keep the separation between getting with `id` and with a filter? 
Also, do you mean they are each optional? I'm asking because since `name` is defined as unique, you are using `First` whereas service_type isn't so you use `Find`

ygalblum:
Also, can you rebase on main. It will resolve the conflicts and remove the duplicated commit

api/v1alpha1/openapi.yaml
ygalblum: I used `v1alpha1` in the path because that's the API version

internal/store/provider.go
gciavarrini: Done. Added ProviderFilter with optional pointer fields. Get(id) stays separate for primary key lookups. `List(filter)` uses `Find` (returns list), `GetByName` uses `First` (unique lookup). WDYT?

internal/store/db.go
gciavarrini: Good catch. Added validation that warns on invalid `DB_TYPE` and defaults to `sqlite`.

internal/store/model/provider.go
gciavarrini: Keeping only the model to core fields (`ID`, `Name`, `ServiceType`, `Endpoint`, and timestamps).

machacekondra (approved):

api/v1alpha1/openapi.yaml
gciavarrini: @machacekondra I see that you marked this as resolved. I don't have a clear answer for your questions TBH. Do we want to start a discussion (in a slack thread?) about this so that i can update the enhancement?

machacekondra: Yep, either we can have slack thread, or note in agenda for meeting. I am fine to have it for now, I was just curios what's reasioning about it.

gciavarrini: added in the agenda

jenniferubah (approved):

internal/store/provider.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+type Provider interface {
+	List(ctx context.Context, filter *ProviderFilter) (model.ProviderList, error)
+	Create(ctx context.Context, provider model.Provider) (*model.Provider, error)
+	Delete(ctx context.Context, id uuid.UUID) error
+	Update(ctx context.Context, provider model.Provider) (*model.Provider, error)
+	Get(ctx context.Context, id uuid.UUID) (*model.Provider, error)
+	GetByName(ctx context.Context, name string) (*model.Provider, error)
+	ExistsByID(ctx context.Context, id uuid.UUID) (bool, error)
```
gabriel-farache: should we also have `ExistsByName`? As far as I remember, both are unique a may serve as "id" so it may be used by the core before giving it to the DB (which should reject the insert into statement if the constraint is violated)

internal/store/provider.go
```diff
@@ -0,0 +1,117 @@
...
+			return nil, ErrProviderNotFound
+		}
+		return nil, err
+	}
+	return &provider, nil
+}
+
+func (s *ProviderStore) GetByName(ctx context.Context, name string) (*model.Provider, error) {
+	var provider model.Provider
+	if err := s.db.WithContext(ctx).Where("name = ?", name).First(&provider).Error; err != nil {
```
gabriel-farache: why not use `.Where(&model.Provider{Name: *filter.Name})` like you did in the LIST (I know that in the LIST it's mandatory as the filter are optional and additive)

internal/store/provider_test.go
```diff
@@ -0,0 +1,202 @@
...
+		ctx = context.Background()
+	})
+
+	AfterEach(func() {
+		sqlDB, _ := db.DB()
+		sqlDB.Close()
+	})
+
+	Describe("Create", func() {
+		It("persists the provider", func() {
```
gabriel-farache, line 41: another test to validate "name" uniqueness may be nice

internal/store/provider_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+		It("returns ErrProviderNotFound for missing ID", func() {
+			err := providerStore.Delete(ctx, uuid.New())
+
+			Expect(err).To(Equal(store.ErrProviderNotFound))
+		})
+	})
+
+	Describe("Update", func() {
+		It("modifies existing provider", func() {
```
gabriel-farache, line 175: you could also test Update on non-existing provider

gabriel-farache (commented):
few comments/questions

internal/store/provider.go
gciavarrini: I'd prefer to keep the interface minimal and avoid adding code we don't strictly need right now. We can rely on `GetByName` and check if it returns `ErrProviderNotFound`. At the moment, we actually need `GetByName` because in the registration flow we need the full provider object. If we need more efficiency in the future, we can add `ExistsByName` then. Does it make sense?

gabriel-farache: > I'd prefer to keep the interface minimal and avoid adding code we don't strictly need right now.

Makes sense, but then, is `ExistsByID` used somewhere? From what I understand this list of method will somehow by mapped to the CRUD API defined by the spec. 
So is this exists method needed and does it needs to be public?

The existence checks is not strictly needed as the SQL query will either return an error or empty row if it does not exist.

internal/store/provider_test.go
gciavarrini: Added. Good catch.

internal/store/provider.go
gciavarrini: You're right, I missed updating these. Fixed both `GetByName` and `ExistsByID`. Thanks for catching it.

internal/store/provider_test.go
gciavarrini: Added. This helped to find a bug! `Update` was using `Save` which silently `upserts`. Now it uses `Updates`  so that both `Delete` and `Update` use the same pattern to detect non-existent providers.
Thanks!

internal/store/provider.go
gciavarrini: argh you're right: it's unused. I checked the references but i missed it's used only in tests. Removing it.

gabriel-farache (approved):

internal/store/model/provider.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type Provider struct {
+	ID          uuid.UUID `gorm:"primaryKey;type:uuid"`
+	Name        string    `gorm:"uniqueIndex;not null"`
+	ServiceType string    `gorm:"column:service_type;not null"`
```
ygalblum: On today's call we said that part of the registration the SP will state the service type api version. Not sure if it's a must for V1, but still worth thinking about this

internal/store/provider.go
ygalblum: Why would you need this function? What's the expected behaviour if there are multiple providers with the same name?

gabriel-farache: the name of the provider is unique as defined in the doc and in the model (https://github.com/gciavarrini/service-provider-api/blob/b03ef8318a1744aee31bd1638d8949edf2747f40/internal/store/model/provider.go#L11) so there can be no name collisions. 

As a user I may want to request the provider by its name as it's unique instead of the UUID

internal/store/model/provider.go
gciavarrini: Good point. 
Updated the enhancement doc to include schemaVersion as a required field in the registration payload (see [enhancements - PR 23](https://github.com/dcm-project/enhancements/pull/23)). 
Also added it to the OpenAPI spec and model in this PR.

internal/store/provider.go
gciavarrini: As @gabriel-farache  said, `name` is unique (enforced by the model). `GetByName` is used in the registration flow to check if a provider with that name already exists and to get the full provider object to update it if it does.

ygalblum: Thanks for the clarification

internal/store/model/provider.go
```diff
@@ -0,0 +1,20 @@
+package model
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type Provider struct {
```
ygalblum, line 9: This model does not include the optional fields of the `Provider` object defined in the OpenAPI spec (`operations`, `metadata` and `status`).
Is that intentional?

gciavarrini: Yes, it's intentional. The team agreed to keep the model minimal and add fields when the business logic needs them (following this discussion https://github.com/dcm-project/service-provider-manager/pull/4#discussion_r1234567). 
The optional fields (operations, metadata, status) will be added when we implement the features that use them.

internal/config/config.go
```diff
@@ -28,5 +30,9 @@ func Load() (*Config, error) {
 	if err := envconfig.Process("", cfg); err != nil {
 		return nil, err
 	}
+	if cfg.Database.Type != "pgsql" && cfg.Database.Type != "sqlite" {
```
machacekondra, line 33: Any idea how we will deploy the DB?

gciavarrini: The DB wiring will be in the next PR (business logic). 
The idea is that the service connects via environment variables (`DB_TYPE`, `DB_HOST`, etc.) and we provision PostgreSQL separately while SQLite is available for local dev.

machacekondra: I thought like, if we will have script to deploy the database here in this project? (I am asking because I've started working on the health check monitoring)

ygalblum: My 2c
- Development and White Box Testing - sqlite
- Black Box Subsystem Testing - Deploy PostgreSQL locally
- E2E Testing and actual deployment -  PostgreSQL is deployed either by the user or as part of the entire solution

gciavarrini: Makes sense. SQLite for dev/unit tests is already working. I'll add a compose file for local PostgreSQL testing in a follow-up.

> PostgreSQL is deployed either by the user or as part of the entire solution

I think letting the user provide their own DB is better, so we don't have the responsibility of managing DB deployment in prod. We can use the compose file just for local dev convenience.

wdyt?

machacekondra: Yeah, actually I was trying to point out the question, if we are the owners of the DB. If yes, we need to make sure we know how to deploy it properly and have DB we support.

machacekondra: > I think letting the user provide their own DB is better, so we don't have the responsibility of managing DB deployment in prod. We can use the compose file just for local dev convenience. wdyt?

I kind of think that's not what user expects.

gabriel-farache: Not sure we should own the DB and all that comes with it.

We may provide some doc to quick install the DB, like an Helm chart or whatever but in the end, the DB will be owned and configured by RHDH (as far as I understand, DCM will be in the RHDH context) or by the user deploying the solution. At least that's how we did it with the Orchestrator.

internal/store/model/provider.go
ygalblum: Not a fan of this view. This PR does not include any logic. So, the same can be said about the mandatory fields.
But, not gonna argue

internal/config/config.go
machacekondra: Let's talk about this, I'll put it on agenda. Note that we agreed that every service has it's own database, so it will be quite a few.

ygalblum: However we do the actual deployment (Helm, Operator, Compose, etc.) there will have to be an optional DB. Either the user provides the connection details, or lets the deployment process to run one.
In any case, this is out of scope of this PR

gabriel-farache: > Note that we agreed that every service has it's own database, so it will be quite a few.

In the Orchestrator example, the plugin re-used the PSQL DB deployed by RHDH. So we can have a single PSQL instance (same as RHDH or a new one) with a DB for each services, that does not "sock" me, but yeah let's settle this during a meeting, it will be more efficient

machacekondra: Added to agenda. Resolving the conversation.

internal/store/model/provider.go
machacekondra: Well, there is a reason behind this. We more or less agreed on enpodint, SPtype, name, id. But we didn't agreed on the rest I would say. Like `totalNode`, `totalCpu`, what is it used for in terms of database or any other custom SP type? Most likely those will change, so why to add them, to make complicated database upgrade scripts? If we would have well defined DB schema, then it would make sense IMHO.

internal/store/model/provider.go
machacekondra: We didn't define a specific DB schema for each service type, so we can't provide all the fields. I think we need a proper design for optional fields per service type. That's why IMHO it doesn't make sense to add those, because we need to change the DB schema often. Better to design first, we can work for now (in v1) with just mandatory fields, which are same for all service types.

ygalblum (approved):

PR #6, opened 2026-01-26
FLPATH-3008: Add business logic and handlers [4/5]

gciavarrini:
Adds the service layer and HTTP handlers for provider registration.

The ProviderService implements idempotent registration as described in the [sp-registration-flow](https://github.com/dcm-project/enhancements/blob/main/enhancements/sp-registration-flow/sp-registration-flow.md) enhancement. Handlers wire everything together and return proper error responses.

2026-01-26-provider-crud

## PR Series

1. OpenAPI spec + code generation
2. Core infrastructure
3. Data layer
4. **This PR** - Business logic + handlers
5. Client library

## Fixes
https://issues.redhat.com/browse/FLPATH-3008

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,26 @@
+# Build stage
+FROM registry.access.redhat.com/ubi9/go-toolset:latest AS builder
+
+WORKDIR /app
+
+# Copy go mod files first for better caching
+COPY go.mod go.sum ./
+RUN go mod download
+
+# Copy source code
+COPY . .
+
+# Build the binary
+USER root
+RUN CGO_ENABLED=0 GOOS=linux go build -buildvcs=false -o service-provider-manager ./cmd/service-provider-manager
+
+# Runtime stage
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+
+WORKDIR /app
+
+COPY --from=builder /app/service-provider-manager .
+
+EXPOSE 8080
+
+ENTRYPOINT ["./service-provider-manager"]
--- a/README.md
+++ b/README.md
@@ -2,6 +2,61 @@
 
 DCM Service Provider Manager - Registration and management of Service Providers.
 
+## Quick Start
+
+### Prerequisites
+
+- Go 1.25+
+- Podman or Docker with Compose
+
+### Run Locally
+
+```bash
+# Start the service with PostgreSQL
+podman-compose up -d
+
+# Check health
+curl http://localhost:8080/api/v1alpha1/health
+```
+
+### Build
+
+```bash
+make build        # Build binary
+make run          # Run locally (requires PostgreSQL)
+```
+
+### Test
+
+```bash
+make test         # Unit tests
+make test-e2e     # E2E tests (requires running services)
+```
+
+### API
+
+| Method | Endpoint | Description |
+|--------|----------|-------------|
+| GET | `/api/v1alpha1/health` | Health check |
+| POST | `/api/v1alpha1/providers` | Register provider (idempotent) |
+| GET | `/api/v1alpha1/providers` | List providers |
+| GET | `/api/v1alpha1/providers/{id}` | Get provider |
+| PUT | `/api/v1alpha1/providers/{id}` | Update provider |
+| DELETE | `/api/v1alpha1/providers/{id}` | Delete provider |
+
+### Configuration
+
+Environment variables:
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `SVC_ADDRESS` | `:8080` | Service listen address |
+| `DB_HOST` | `localhost` | PostgreSQL host |
+| `DB_PORT` | `5432` | PostgreSQL port |
+| `DB_NAME` | `service-provider` | Database name |
+| `DB_USER` | *(none)* | Database user (required for pgsql) |
+| `DB_PASS` | *(none)* | Database password (required for pgsql) |
+
 ## License
 
 Apache 2.0 - see [LICENSE](LICENSE)
--- a/README.md
+++ b/README.md
@@ -54,8 +54,8 @@ Environment variables:
 | `DB_HOST` | `localhost` | PostgreSQL host |
 | `DB_PORT` | `5432` | PostgreSQL port |
 | `DB_NAME` | `service-provider` | Database name |
-| `DB_USER` | `admin` | Database user |
-| `DB_PASS` | `adminpass` | Database password |
+| `DB_USER` | *(none)* | Database user (required for pgsql) |
+| `DB_PASS` | *(none)* | Database password (required for pgsql) |
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -54,8 +54,8 @@ Environment variables:
 | `DB_HOST` | `localhost` | PostgreSQL host |
 | `DB_PORT` | `5432` | PostgreSQL port |
 | `DB_NAME` | `service-provider` | Database name |
-| `DB_USER` | *(none)* | Database user (required for pgsql) |
-| `DB_PASS` | *(none)* | Database password (required for pgsql) |
+| `DB_USER` | `admin` | Database user |
+| `DB_PASS` | `adminpass` | Database password |
 
 ## License
 
--- a/internal/api_server/server.go
+++ b/internal/api_server/server.go
@@ -3,10 +3,12 @@ package apiserver
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
 	"net/http"
 	"time"
 
+	"github.com/dcm-project/service-provider-manager/api/v1alpha1"
 	"github.com/dcm-project/service-provider-manager/internal/api/server"
 	"github.com/dcm-project/service-provider-manager/internal/config"
 	"github.com/go-chi/chi/v5"
@@ -34,7 +36,15 @@ func (s *Server) Run(ctx context.Context) error {
 	router.Use(middleware.Logger)
 	router.Use(middleware.Recoverer)
 
-	server.HandlerFromMux(server.NewStrictHandler(s.handler, nil), router)
+	swagger, err := v1alpha1.GetSwagger()
+	if err != nil {
+		return fmt.Errorf("load OpenAPI spec: %w", err)
+	}
+	if len(swagger.Servers) == 0 {
+		return fmt.Errorf("OpenAPI spec missing servers configuration")
+	}
+
+	server.HandlerFromMuxWithBaseURL(server.NewStrictHandler(s.handler, nil), router, swagger.Servers[0].URL)
 
 	srv := http.Server{Handler: router}
 
--- a/internal/handlers/handler.go
+++ b/internal/handlers/handler.go
@@ -1,0 +1,111 @@
+package handlers
+
+import (
+	"context"
+
+	"github.com/dcm-project/service-provider-manager/internal/api/server"
+	"github.com/dcm-project/service-provider-manager/internal/service"
+)
+
+// Handler implements the generated StrictServerInterface for the Provider API.
+type Handler struct {
+	providerService *service.ProviderService
+}
+
+// NewHandler creates a new Handler with the given provider service.
+func NewHandler(providerService *service.ProviderService) *Handler {
+	return &Handler{providerService: providerService}
+}
+
+// Ensure Handler implements StrictServerInterface
+var _ server.StrictServerInterface = (*Handler)(nil)
+
+func (h *Handler) GetHealth(ctx context.Context, request server.GetHealthRequestObject) (server.GetHealthResponseObject, error) {
+	status := "ok"
+	path := "health"
+	return server.GetHealth200JSONResponse{Status: &status, Path: &path}, nil
+}
+
+func (h *Handler) ListProviders(ctx context.Context, request server.ListProvidersRequestObject) (server.ListProvidersResponseObject, error) {
+	var serviceType string
+	if request.Params.Type != nil {
+		serviceType = *request.Params.Type
+	}
+
+	providers, err := h.providerService.ListProviders(ctx, serviceType)
+	if err != nil {
+		return server.ListProviders400ApplicationProblemPlusJSONResponse(newError("list-error", "Failed to list providers", err.Error(), 400)), nil
+	}
+
+	return server.ListProviders200JSONResponse{Providers: &providers}, nil
+}
+
+func (h *Handler) CreateProvider(ctx context.Context, request server.CreateProviderRequestObject) (server.CreateProviderResponseObject, error) {
+	response, err := h.providerService.RegisterProvider(ctx, request.Body, request.Params.Id)
+	if err != nil {
+		if svcErr, ok := err.(*service.ServiceError); ok {
+			switch svcErr.Code {
+			case service.ErrCodeValidation:
+				return server.CreateProvider400ApplicationProblemPlusJSONResponse(newError("validation-error", "Validation failed", svcErr.Message, 400)), nil
+			case service.ErrCodeConflict:
+				return server.CreateProvider409ApplicationProblemPlusJSONResponse(newError("conflict", "Resource conflict", svcErr.Message, 409)), nil
+			}
+		}
+		return server.CreateProvider400ApplicationProblemPlusJSONResponse(newError("create-error", "Failed to create provider", err.Error(), 400)), nil
+	}
+
+	if response.Status != nil && *response.Status == server.Updated {
+		return server.CreateProvider200JSONResponse(*response), nil
+	}
+	return server.CreateProvider201JSONResponse(*response), nil
+}
+
+func (h *Handler) GetProvider(ctx context.Context, request server.GetProviderRequestObject) (server.GetProviderResponseObject, error) {
+	provider, err := h.providerService.GetProvider(ctx, request.ProviderId.String())
+	if err != nil {
+		if svcErr, ok := err.(*service.ServiceError); ok && svcErr.Code == service.ErrCodeNotFound {
+			return server.GetProvider404ApplicationProblemPlusJSONResponse(newError("not-found", "Provider not found", svcErr.Message, 404)), nil
+		}
+		return server.GetProvider400ApplicationProblemPlusJSONResponse(newError("get-error", "Failed to get provider", err.Error(), 400)), nil
+	}
+
+	return server.GetProvider200JSONResponse(*provider), nil
+}
+
+func (h *Handler) ApplyProvider(ctx context.Context, request server.ApplyProviderRequestObject) (server.ApplyProviderResponseObject, error) {
+	provider, err := h.providerService.UpdateProvider(ctx, request.ProviderId.String(), request.Body)
+	if err != nil {
+		if svcErr, ok := err.(*service.ServiceError); ok {
+			switch svcErr.Code {
+			case service.ErrCodeNotFound:
+				return server.ApplyProvider404ApplicationProblemPlusJSONResponse(newError("not-found", "Provider not found", svcErr.Message, 404)), nil
+			case service.ErrCodeConflict:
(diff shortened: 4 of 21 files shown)


internal/api_server/server.go
```diff
@@ -34,7 +34,7 @@ func (s *Server) Run(ctx context.Context) error {
 	router.Use(middleware.Logger)
 	router.Use(middleware.Recoverer)
 
-	server.HandlerFromMux(server.NewStrictHandler(s.handler, nil), router)
+	server.HandlerFromMuxWithBaseURL(server.NewStrictHandler(s.handler, nil), router, "/api/v1alpha1")
```
ygalblum: Consider getting the base URL from the OpenAPI file.
See example here: https://github.com/dcm-project/policy-manager/blob/c51cbaeeb125e8b783b4ab81401b9fbd66af2502/internal/apiserver/server.go#L43

internal/handlers/handler.go
```diff
@@ -0,0 +1,110 @@
...
+	providerService *service.ProviderService
+}
+
+// NewHandler creates a new Handler with the given provider service.
+func NewHandler(providerService *service.ProviderService) *Handler {
+	return &Handler{providerService: providerService}
+}
+
+// Ensure Handler implements StrictServerInterface
+var _ server.StrictServerInterface = (*Handler)(nil)
```
ygalblum, line 21: Not sure I understand why is this needed? Is this for compilation check? Alternatively, can `NewHandler` return `*server.StrictServerInterface` instead of `*Handler`? This will do the same, no?

internal/handlers/handler.go
```diff
@@ -0,0 +1,110 @@
...
+	return server.GetHealth200JSONResponse{Status: &status}, nil
+}
+
+func (h *Handler) ListProviders(ctx context.Context, request server.ListProvidersRequestObject) (server.ListProvidersResponseObject, error) {
+	var serviceType string
+	if request.Params.Type != nil {
+		serviceType = *request.Params.Type
+	}
+
+	providers, err := h.providerService.ListProviders(ctx, serviceType)
```
ygalblum: This implementation is missing pagination. The max size and token should either be unpacked here and passed to the `ProviderService` as offset and limit or just pass them as is and let it unwrap them. 
Note that you also don't return the token for the next call in the returned `server.ListProvidersResponseObject`

internal/handlers/handler_test.go
```diff
@@ -0,0 +1,169 @@
...
+			resp, err := handler.GetHealth(ctx, server.GetHealthRequestObject{})
+
+			Expect(err).NotTo(HaveOccurred())
+			jsonResp, ok := resp.(server.GetHealth200JSONResponse)
+			Expect(ok).To(BeTrue())
+			Expect(*jsonResp.Status).To(Equal("ok"))
+		})
+	})
+
+	Describe("CreateProvider", func() {
```
ygalblum, line 57: Missing tests
- Returning 200
- Conflict 409

internal/handlers/handler_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+			resp, err := handler.CreateProvider(ctx, req)
+
+			Expect(err).NotTo(HaveOccurred())
+			_, ok := resp.(server.CreateProvider201JSONResponse)
+			Expect(ok).To(BeTrue())
+		})
+	})
+
+	Describe("ListProviders", func() {
```
ygalblum, line 131: Missing tests
- Simple success
- Pagination

internal/handlers/handler_test.go
```diff
@@ -0,0 +1,169 @@
...
+			resp, err := handler.ListProviders(ctx, req)
+
+			Expect(err).NotTo(HaveOccurred())
+			jsonResp, ok := resp.(server.ListProviders200JSONResponse)
+			Expect(ok).To(BeTrue())
+			Expect(*jsonResp.Providers).To(BeEmpty())
+		})
+	})
+
+	Describe("GetProvider", func() {
```
ygalblum, line 167: Missing simple success test

internal/handlers/handler_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+			resp, err := handler.ApplyProvider(ctx, req)
+
+			Expect(err).NotTo(HaveOccurred())
+			_, ok := resp.(server.ApplyProvider404ApplicationProblemPlusJSONResponse)
+			Expect(ok).To(BeTrue())
+		})
+	})
+
+	Describe("DeleteProvider", func() {
```
ygalblum, line 259: Missing successful delete test

internal/service/provider.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+		return nil, err
+	}
+
+	return ModelToProvider(provider), nil
+}
+
+// ListProviders returns all providers, optionally filtered by service type.
+func (s *ProviderService) ListProviders(ctx context.Context, serviceType string) ([]server.Provider, error) {
+	var filter *store.ProviderFilter
```
ygalblum, line 187: Similar to the comment in the handler, pagination should be handled somewhere in this flow

internal/service/provider.go
```diff
@@ -0,0 +1,200 @@
...
+		other, err := s.store.Provider().GetByName(ctx, update.Name)
+		if err != nil && !errors.Is(err, store.ErrProviderNotFound) {
+			return nil, err
+		}
+		if other != nil && other.ID != id {
+			return nil, &ServiceError{Code: ErrCodeConflict, Message: fmt.Sprintf("name '%s' is already taken", update.Name)}
+		}
+	}
+
+	existing.Name = update.Name
```
ygalblum: Can `updateExistingProvider` be reused here? If I read this correctly, the only missing line is the one setting the `Name` and I think it can be copied there as well

Containerfile
```diff
@@ -0,0 +1,27 @@
+# Build stage
```
ygalblum, line 1: Let's use UBI based images.
`registry.access.redhat.com/ubi9/go-toolset` for builder
`registry.access.redhat.com/ubi9/ubi-minimal` for runtime

internal/handlers/handler.go
```diff
@@ -0,0 +1,110 @@
...
+func NewHandler(providerService *service.ProviderService) *Handler {
+	return &Handler{providerService: providerService}
+}
+
+// Ensure Handler implements StrictServerInterface
+var _ server.StrictServerInterface = (*Handler)(nil)
+
+func (h *Handler) GetHealth(ctx context.Context, request server.GetHealthRequestObject) (server.GetHealthResponseObject, error) {
+	status := "ok"
+	return server.GetHealth200JSONResponse{Status: &status}, nil
```
machacekondra: Gloaria, I think this line is not making AEP happy :D

internal/service/provider.go
```diff
@@ -0,0 +1,200 @@
...
+	var providedID *uuid.UUID
+	if req.Id != nil {
+		id := uuid.UUID(*req.Id)
+		providedID = &id
+	} else if queryID != nil {
+		id := uuid.UUID(*queryID)
+		providedID = &id
+	}
+
+	// Check if provider with this name already exists
```
gabriel-farache: should you also check that if the ID was provided, it does not already exist?

gabriel-farache: ah, I see the check is done after, could you maybe extract those checks into dedicated methods like `CheckProvidedProviderNameExists` and `CheckProvidedIdExists`

And why not have the check for `providedID` right after the `if` statements above?

internal/service/provider.go
```diff
@@ -0,0 +1,200 @@
...
+	existing, err := s.store.Provider().GetByName(ctx, req.Name)
```
gabriel-farache: `existing` and `exists` naming are confusing and does not help with the reading
Suggestion: renames to `existingFromName` and `existingFromId` or something similar

internal/service/provider.go
```diff
@@ -0,0 +1,200 @@
...
+	// Name doesn't exist - create new provider
+	var providerID uuid.UUID
+	if providedID != nil {
+		providerID = *providedID
+		// Check if this ID is already taken by another provider
+		exists, err := s.store.Provider().ExistsByID(ctx, providerID)
+		if err != nil {
+			return nil, err
+		}
+		if exists {
```
gabriel-farache: IIUC, here, if the provider exists but has not the same name as the one provided, we exit with error, is that correct?
But is it really what we want? Why not update the name?

internal/service/provider.go
```diff
@@ -0,0 +1,200 @@
...
+func NewProviderService(store store.Store) *ProviderService {
+	return &ProviderService{store: store}
+}
+
+// RegisterProvider implements idempotent provider registration per the DCM spec.
+// Returns status "registered" for new providers, "updated" for existing ones.
+// Returns ErrCodeConflict if name exists with different ID or ID exists with different name.
+func (s *ProviderService) RegisterProvider(ctx context.Context, req *server.Provider, queryID *openapi_types.UUID) (*server.Provider, error) {
+	// Determine if caller provided an ID
+	var providedID *uuid.UUID
```
gabriel-farache: oh god `providedID` and `providerID` are so close it was not understanding what was happening, once I noticed the damn letter, I started to understand :)

Would you be OK to rename one or both of them? Maybe `suppliedId` and `id`?

internal/service/provider.go
gabriel-farache: ok, now I understand better: if the name exists, then you check if the supplied ID matched the ID of the provider with the supplied name and if it's the case, you update the provider, otherwise you fail as it means the name is already taken which violates the uniqueness constraint.

Could this check occurs in the caller instead of here? 
Or rename the method's name to `RegisterOrUpdate` to make t clear what the method will be trying to do

internal/service/provider_test.go
```diff
@@ -0,0 +1,250 @@
...
+			req.Id = resp1.Id
+			req.Endpoint = "https://updated.example.com"
+			resp2, err := providerService.RegisterProvider(ctx, req, nil)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp2.Status).NotTo(BeNil())
+			Expect(*resp2.Status).To(Equal(server.Updated))
+		})
+
+		It("updates existing provider with same name and no ID (idempotent)", func() {
```
gabriel-farache, line 71: in case there is no `name` provided, you are relying on the upper layer to throw an error before it reaches the service, correct?

internal/service/provider_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+			Expect(err).To(HaveOccurred())
+			svcErr, ok := err.(*service.ServiceError)
+			Expect(ok).To(BeTrue())
+			Expect(svcErr.Code).To(Equal(service.ErrCodeNotFound))
+		})
+	})
+
+	Describe("ListProviders", func() {
+		It("returns all providers", func() {
```
gabriel-farache, line 140: you will need test(s) for pagination I think or at least the concrete translation from token pagination to SQL limit and offset

internal/store/provider.go
```diff
@@ -110,3 +111,11 @@ func (s *ProviderStore) GetByName(ctx context.Context, name string) (*model.Prov
 	}
 	return &provider, nil
 }
+
+func (s *ProviderStore) ExistsByID(ctx context.Context, id uuid.UUID) (bool, error) {
+	var count int64
+	if err := s.db.WithContext(ctx).Model(&model.Provider{}).Where(&model.Provider{ID: id}).Count(&count).Error; err != nil {
```
gabriel-farache: why not use `exists` instead of counting?

internal/handlers/handler.go
gciavarrini: Yes, it's a compile-time check. The pattern `(var _ Interface = (*Type)(nil))` is standard in Go and it catches missing methods with a clear error.
Returning the interface from `NewHandler` works, but I followed the "accept interfaces, return concrete types" [guideline](https://go.dev/wiki/CodeReviewComments#interfaces)

README.md
```diff
@@ -2,6 +2,61 @@
...
+Environment variables:
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `SVC_ADDRESS` | `:8080` | Service listen address |
+| `DB_HOST` | `localhost` | PostgreSQL host |
+| `DB_PORT` | `5432` | PostgreSQL port |
+| `DB_NAME` | `service-provider` | Database name |
+| `DB_USER` | `admin` | Database user |
+| `DB_PASS` | `adminpass` | Database password |
```
gabriel-farache: We already talk about that with @ygalblum in one of his PR but I have trouble understanding why we should have default values for DB username and password. Why have default values for that? What is the advantage of doing that? (these are real questions, I want to understand why it's good to do that)

internal/handlers/handler_test.go
gciavarrini: Added a simple success test for ListProviders. Pagination logic test is covered by ([provider_test.go](https://github.com/gciavarrini/service-provider-api/blob/58f9051280231fef9b9c25ba0770ed804b48c1c1/internal/service/provider_test.go#L165C8-L217))

internal/handlers/handler_test.go
gciavarrini: Added

internal/handlers/handler_test.go
gciavarrini: Added

internal/service/provider.go
gciavarrini: Addressed in commit 7555af2.

internal/service/provider.go
gciavarrini: Good catch! Refactored

internal/handlers/handler.go
gciavarrini: Funny, I flagged the same issue in your PR but was completely blind to it in my own code 😅 Added the missing path field.

Containerfile
gciavarrini: Switched to UBI images. Also lowered go.mod to 1.25.5 to match go-toolset.

internal/service/provider.go
gciavarrini: You can update the name. The error only happens if the new name is already taken by another service provider (we have the name uniqueness constraint). If the name is available, the update proceeds and the name is changed. Makes sense?

machacekondra (approved):

internal/service/provider.go
gciavarrini: You're right. Renamed `providedID` to `requestedID`.

internal/service/provider.go
gciavarrini: Refactored with helper methods `checkNameConflict()` and `resolveProviderID()`

internal/service/provider.go
gciavarrini: Renamed to `RegisterOrUpdateProvider`.

Regarding moving checks to the caller I think keeping them in the service layer let the idempotent registration logic to stay in the same place and can be reused. 
For example, if we'll add a CLI tool to register providers, it calls `RegisterOrUpdateProvider()` and gets the same idempotent behavior without duplicating the conflict checks.

internal/service/provider_test.go
gciavarrini: Correct. 
The `name` field is marked as `required` in the OpenAPI schema. For this reasont the generated server rejects requests without `name` before. This avoids duplicating validation

internal/service/provider_test.go
gciavarrini: Addressed in commit 7555af2

README.md
gciavarrini: I think you refer to this [thread](https://github.com/dcm-project/policy-manager/pull/4#discussion_r2728302970) in policy-manager PR.

The defaults exist for developers convenience so that when running the service directly (outside compose). 

Without defaults:
```go
# Developer would need:
DB_USER=admin DB_PASS=adminpass go run ./cmd/service-provider-manager
```

The `compose.yaml` already sets these explicitly, so `podman-compose up` works.

I can remove the defaults if we prefer to fail explicitly when credentials aren't set. WDYT?

internal/store/provider.go
gciavarrini: you're right. AI suggested `Take()` as the GORM-idiomatic approach, which stops at the first match. Updated

internal/service/provider.go
gabriel-farache: not sure I follow so I'll try to put the workflow execution here:
1. check name exists in DB
2. if exists, check if same ID and update (success), otherwise throw error (fail)
3. => name does not exists in DB yet
4. check ID exists in DB
5. if ID exists, throw error (fail), otherwise create the provider in the DB

Did I get that right or did I miss something?
If I get it right, if the ID exists why should we fail instead of updating the other fields like we did for the name? Or are we failing because the name is required in the payload for the POST which mean the SP is trying to register a new entry instead of updating it? It would be like an update hijack.

Ok, I guess I get it now, but the in-place update in case of name existence but not for ID is confusing without the upper layers context

internal/api_server/server.go
gciavarrini: Thanks! Updated.
I also added a check for empty servers so if the spec gets misconfigured, we'll catch a malformed spec early.

README.md
gabriel-farache: I see your point of dev experience with the defaults. However, is it right to set defaults that will be shipped to PROD for DEV convenience? Especially that this could be overcome by creating a target in the Makefile that are matching the credentials defined in the compose file.

This is not a blocker so I will approve the PR to allow you to continue the work, we can update that in all repos later if needs be.

What are @dcm-project/team-dcm thinking about that?

gabriel-farache (approved):

internal/service/provider.go
gciavarrini: > Did I get that right or did I miss something?

Yes, correct

> if the ID exists why should we fail instead of updating the other fields like we did for the name? 

We fail on existing ID because name is the natural key for matching (as per the enhancement and spec). When a provider re-registers, it uses the same name to say "it's me again." A new name + existing ID means someone is trying to claim another provider's identity.

> Or are we failing because the name is required in the payload for the POST which mean the SP is trying to register a new entry instead of updating it? It would be like an update hijack.

Since name is required, providing a new name signals "I'm a new provider." If the ID already belongs to someone else, that's a conflict.

internal/handlers/handler.go
ygalblum: OK, thanks for the explanation

internal/service/provider.go
```diff
@@ -0,0 +1,288 @@
...
+	if queryID != nil {
+		id := uuid.UUID(*queryID)
+		return &id
+	}
+	return nil
+}
+
+// checkNameConflict returns the existing provider if name exists and is valid for update.
+// Returns ErrCodeConflict if name exists with a different ID than requested.
+func (s *ProviderService) checkNameConflict(ctx context.Context, name string, requestedID *uuid.UUID) (*model.Provider, error) {
```
ygalblum: Sorry for the nit picking here. But, the name of the method is a bit misleading. Yes, it checks for name conflict, but is also (and maybe mostly) fetches and returns the model and which is used by the caller.

README.md
gciavarrini: I see your point, it's the safer approach. 
I was too focused on dev convenience. Removed the defaults for `DB_USER` and `DB_PASS`. For Postgres, the connection will fail if credentials aren't set. SQLite mode doesn't need them.
For local dev, `compose.yaml` sets the env vars so `podman-compose up` still works out of the box.

internal/service/provider.go
gciavarrini: You're right, I'm terrible in picking decent method names :)
Renamed to `findExistingByName`. WDYT?

ygalblum (approved):

internal/service/provider.go
ygalblum: 👍🏼

jenniferubah (approved):

gabriel-farache:
@gciavarrini can you clean the commits so we can merge the PR? :)

PR #16, opened 2026-01-30
Add SPs health check monitoring

machacekondra:


api/v1alpha1/openapi.yaml
```diff
@@ -113,13 +113,13 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Provider'
+                $ref: '#/components/schemas/ProviderResponse'
```
ygalblum: I don't know why the AEP linter is not catching this. But, these changes do not adhere to the spec:
- The returned payload for the create operation should be of the same type as the request body
- Both the `Provider` and the `ProviderResponse` have the same `x-aep-resource` struct

Why not just add `health_status` to the `Provider` object and mark it as `readOnly`?

cmd/service-provider-manager/main.go
```diff
@@ -45,6 +46,12 @@ func main() {
 	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer cancel()
 
+	// Start health check monitor
+	healthMonitor := healthcheck.NewMonitor(dataStore.Provider(), cfg.HealthCheck)
+	healthMonitor.Start(ctx)
+	defer healthMonitor.Stop()
+
+	log.Printf("Health check monitor started (interval: 10s)")
```
ygalblum, line 53: Sorry for the nit picking here. But, this print belong before the empty line

api/v1alpha1/openapi.yaml
machacekondra: Oh, didn't know this is AEP thing, but make total sense to make it readonly, and not have two objects. Thanks!

ygalblum (approved):

internal/store/provider.go
```diff
@@ -156,3 +161,30 @@ func (s *ProviderStore) ExistsByID(ctx context.Context, id uuid.UUID) (bool, err
...
+		Find(&providers).Error; err != nil {
+		return nil, err
+	}
+	return providers, nil
+}
+
+// UpdateHealthStatus updates the health status and tracking fields for a provider.
+func (s *ProviderStore) UpdateHealthStatus(ctx context.Context, id uuid.UUID, status model.HealthStatus, consecutiveFailures int, nextCheck time.Time) error {
+	result := s.db.WithContext(ctx).Model(&model.Provider{}).Where("id = ?", id).Updates(map[string]interface{}{
+		"health_status":        status,
```
jenniferubah, line 179: Are we missing unit tests these two functions in the store provider test:https://github.com/dcm-project/service-provider-manager/blob/main/internal/store/provider_test.go?

jenniferubah (approved):
Looks good to me.
Just missing provider store tests

PR #17, opened 2026-01-30
Add Data Store for Resource Manager

jenniferubah:
Add data store layer and unit tests for SP Resource Manager

Assisted By: Cursor (claude-4.5-opus-high)


internal/store/model/service_type_instance.go
```diff
@@ -0,0 +1,21 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"gorm.io/datatypes"
+)
+
+type ServiceTypeInstance struct {
+	ID           uuid.UUID      `gorm:"primaryKey;type:uuid"`
+	ProviderName string         `gorm:"column:provider_name;not null"`
+	ServiceType  string         `gorm:"column:service_type;not null"`
```
ygalblum: I guess I missed that in the OpenAPI Spec, should it also keep the `ServiceTypeVersion`?

jenniferubah: Not sure, the `ServiceTypeVersion` is inside the spec field so do we need to extract/add it on a column level?

gabriel-farache: I remember we had a discussion in another PR about having both `serviceType` and `serviceTypeVersion` (not sure if it was for "service type" or another resource) inside `spec` or at the top level of the payload. But both were at the same level.

That being said, do we really need the version in the resource manager? If I remember correctly, the Placement Manager will holds the CatalogItem information and will the service that request the instance creation to the Resource Manager. The Resource Manager will kind-of be a private service in the end no? 
When any external user/program want to interact with DCM, it will do it through others services (like the Placement Manager or even the Catalog Manager) no?

In the `spec` there can be anything, any data that should match the schema version of the service type I think.

internal/store/resource_manager/service_instance_test.go
gabriel-farache, line 1: Why not use ginko as done in other managers (e.g: https://github.com/dcm-project/policy-manager/blob/main/internal/store/policy_test.go)

jenniferubah: Updated to use ginkgo to be consistent across

internal/store/model/service_type_instance.go
jenniferubah: Let's discuss this during the meeting tomorrow. Added it to agenda.

ygalblum: The reason we said we'll put the `serviceType` in the root was to allow it to verify that the chosen SP does in fact support it. However, since SPs also register the `serviceTypeVersion`, the latter is required for the validation. In the current definition, the `spec` is undefined. So, nothing mandates these fields.
We could also remove these fields and skip the validation altogether and have the SP fail the request if the service type and version are not supported

jenniferubah: You're right, we now need this field as well for validation. We can discuss whether we want to keep or skip the validation during the meeting.

jenniferubah: As agreed, updated the [ADR](https://github.com/dcm-project/enhancements/pull/27) and code (see here [e547897](https://github.com/dcm-project/service-provider-manager/pull/17/commits/e5478975b8890b89a9dbf376ad6cd5760fca4539)) to remove the serviceType field and skip the validation ( to be handled by Policy) within SPRM

internal/store/model/service_type_instance.go
```diff
@@ -0,0 +1,20 @@
+package model
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+	"gorm.io/datatypes"
+)
+
+type ServiceTypeInstance struct {
```
machacekondra, line 10: What is the difference between `ServiceTypeInstance` and `CatalogItemInstance`?

jenniferubah: The `CatalogItemInstance` references the catalog item and aslo contains sevicetype spec data while the `ServiceTypeInstance` is supposed to be an instance of the servicetype spec and also contains the selected provider (righ now, main difference is the payload structure)

gciavarrini (approved):
lgtm

internal/store/model/service_type_instance.go
gabriel-farache: If I remember correctly, we introduced the "service type instances" based on the original figma which was listing all resources managed by DCM. At that time, I am not sure we were considering the catalog item nor catalog item instances in the big picture.

Because, yes the service type instance is a running resource for a specific type. But the users are using catalog items, not service type when requesting a resources to be provisioned. What contains the catalog item instances apart from a reference to the service type instance (ignoring catalog item reference as well)? Should those instances be merged into 1? Do we want to keep those layers to have 1:1 mapping?

jenniferubah: The `ServiceTypeInstance` will also have the status and may evolve to have the configuration for the declared resource. I don't think/know that we want to merge those two.

ygalblum: @gabriel-farache you are correct, the catalog was not really defined in that figma.
The idea is that first there's a ServiceType which is a definition of the parameters needed to provision an instance of a certain service type. As can be understood here, the specific instance of such a service is a ServiceTypeInstance. 

We could have stay here and maybe that would have been simpler to understand and implement. However, it was decided that users are not supposed to directly create ServiceTypeInstances. Instead, admins will define CatalogItem which wrap the parameters of a ServiceType (by either setting or limiting the values). Users will then create `CatalogItemInstances`. 

However, SPs were designed to work with `ServiceTypeInstance`. So, the `CatalogManager` is in charge of taking a `CatalogItemInstance` and build a `ServiceTypeInstance` and pass it along.

ygalblum (approved):

gabriel-farache (approved):

PR #18, opened 2026-02-04
SPRM CRUD Logic Implementation

jenniferubah:
Service Provider Resource Manager implementatation

- Added service type instance handlers, converter, and error mapping
- Implemented instance service and converter
- Added handler and service unit tests
- Added status field to OpenAPI Spec

Assisted-by: Cursor (Claude Sonnet  4.5)


api/v1alpha1/resource_manager/openapi.yaml
```diff
@@ -244,6 +244,11 @@ components:
           type: string
           description: Name of the provider
           example: "kubevirt-123"
+        status:
+          type: string
+          readOnly: true
+          description: Status of the instance
+          example: "READY"
```
jenniferubah, line 257: Just a note: These will be updated in future PRs once we DCM status mapping implemented

internal/handlers/resource_manager/errors.go
```diff
@@ -0,0 +1,85 @@
...
+
+// handleCreateInstanceError converts a service error to a CreateInstance response.
+func handleCreateInstanceError(err error) server.CreateInstanceResponseObject {
+	var svcErr *service.ServiceError
+	if errors.As(err, &svcErr) {
+		switch svcErr.Code {
+		case service.ErrCodeValidation:
+			return server.CreateInstance400ApplicationProblemPlusJSONResponse(newError("validation-error", "Validation failed", svcErr.Message, 400))
+		case service.ErrCodeNotFound:
+			return server.CreateInstance400ApplicationProblemPlusJSONResponse(newError("not-found", "Resource not found", svcErr.Message, 400))
```
gciavarrini: Not found error should be `404`.
Also i suggest adding `404` response in the `openapi.yaml` to the `create-instance` operationId

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,289 @@
...
+			}
+		}
+		return nil, err
+	}
+
+	// Check Provider if provider is not in ready state
+	if provider.HealthStatus != model.HealthStatusReady {
+		return nil, &service.ServiceError{
+			Code:    service.ErrCodeProviderError,
+			Message: fmt.Sprintf("provider is not in ready state: %v", err),
```
gciavarrini: the error message uses `err`, but at that point `err` is `nil` if provider lookup succeeded or is the old value from an earlier step. So, the message will show "provider is not in ready state: <nil>" or an unrelated error. I think we should use the provider’s health status instead:

```suggestion
			Message: fmt.Sprintf("provider '%s' is not in ready state (current status: %s)", providerName, provider.HealthStatus),
```

internal/handlers/resource_manager/handler.go
```diff
@@ -0,0 +1,86 @@
...
+// ListInstances returns a paginated list of service type instances.
+func (h *Handler) ListInstances(ctx context.Context, request server.ListInstancesRequestObject) (server.ListInstancesResponseObject, error) {
+	var pageToken string
+	if request.Params.PageToken != nil {
+		pageToken = *request.Params.PageToken
+	}
+
+	result, err := h.instanceService.ListInstances(
+		ctx,
+		request.Params.Type,
```
gciavarrini: There is something wrong here.

`ListInstances` method is defined as
`func (s *InstanceService) ListInstances(ctx context.Context, providerName *string, maxPageSize *int, pageToken string) (*resource_manager.ServiceTypeInstanceList, error) {` in [service_type_instance.go#L129](https://github.com/jenniferubah/service-provider-manager/blob/c65859f7c2451eb925c2f1b6399eb9824c157079/internal/service/resource_manager/service_type_instance.go#L129). So the first argument is the `providerName`

In [resource manager openapi.yaml](https://github.com/jenniferubah/service-provider-manager/blob/c65859f7c2451eb925c2f1b6399eb9824c157079/api/v1alpha1/resource_manager/openapi.yaml#L47-L51) you have `type` and in this line you pass `request.Params.Type`.

Can you double check?

internal/handlers/resource_manager/convert.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+}
+
+// convertAPIToServer converts an API ServiceTypeInstance to a server ServiceTypeInstance.
+func convertAPIToServer(src *resource_manager.ServiceTypeInstance) server.ServiceTypeInstance {
+	return server.ServiceTypeInstance{
+		Id:           src.Id,
+		Path:         src.Path,
+		ProviderName: src.ProviderName,
+		Spec:         src.Spec,
```
gciavarrini, line 24: I know tyou wrote 
"Just a note: These will be updated in future PRs once we DCM status mapping implemented" in this [comment](https://github.com/dcm-project/service-provider-manager/pull/18/changes#r2771378219). Is this referring to the missing 
```go
Missing Status:       src.Status
```
 here?
If so, what do you this to add that mapping as a comment? so it's clear it's intentionally left out

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,289 @@
...
+		ID:           instanceID,
+		ProviderName: providerName,
+		Status:       providerResponse.Status,
+		Spec:         datatypes.JSON(specJSON),
+	}
+
+	created, err := s.store.ServiceTypeInstance().Create(ctx, instance)
+	if err != nil {
+		return nil, err
+	}
```
gabriel-farache, line 79: here we may have an issue: the resource was created by the provider as we get a valid response from the `s.sendToProvider()` call. So if the DB insert fails, we will have some orphaned resource running in the SP, consuming resources for nothing.
I think we should have some tries before returning the error. And the error message should contains the fact that the resource was created by the SP but the issue is internal. The ID of the resource should be provided in the error message as well there can be some action taken to resolve the discrepancy (maybe manual insert later??)

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,289 @@
...
+
+	log.Printf("Created instance: %s for provider: %s", created.ID, providerName)
```
gabriel-farache: this should be after `s.sendToProvider()` 
here it should be "Inserted instance into DB" or something like that

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,289 @@
...
+			// Log error but continue with local deletion
+			log.Printf("Warning: failed to delete instance %s from provider: %v", instanceID, err)
+		}
+	}
+
+	// Delete from database
+	err = s.store.ServiceTypeInstance().Delete(ctx, id)
+	if err != nil {
+		return err
+	}
```
gabriel-farache, line 181: same comment as when the service sends the creation request to the SP successfully

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,289 @@
...
+
+	log.Printf("Deleted instance: %s", instanceID)
```
gabriel-farache: same comment as when the service sends the creation request to the SP successfully: the log message should be after successful deletion and here it should be about cleaning DB

BTW, do we want to have a routine checking for discrepancies between SP and internal DB? Maybe it will be part of the status monitoring PR as the SP will be sending updates about its resources periodically.

internal/service/resource_manager/convert.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+
+	"github.com/dcm-project/service-provider-manager/api/v1alpha1/resource_manager"
+	"github.com/dcm-project/service-provider-manager/internal/service"
+	"github.com/dcm-project/service-provider-manager/internal/store/model"
+)
+
+// ProviderResponse represents the response from a provider during instance creation.
+type ProviderResponse struct {
+	ID     string `json:"id"`
```
gciavarrini, line 13: CAn you double check if `id` is ok or if we need to use `instanceId`? I'm always unsure

internal/service/resource_manager/convert.go
```diff
@@ -0,0 +1,35 @@
...
+	Status string `json:"status"`
```
gciavarrini, line 14: No `metadata` needed here?

gabriel-farache (commented):
just something to take care of in case the failure occurs when interacting with local DB after successful request from SP

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,289 @@
...
+	if err != nil && !errors.Is(err, store.ErrProviderNotFound) {
+		return err
+	}
+
+	// Send delete request to provider if provider still exists
+	if provider != nil {
+		err = s.sendDeleteToProvider(ctx, provider.Endpoint, instanceID)
+		if err != nil {
+			// Log error but continue with local deletion
+			log.Printf("Warning: failed to delete instance %s from provider: %v", instanceID, err)
```
gciavarrini: If the provider delete fails, the code logs a warning and deletes the record from the DB. That can leave resources orphaned on the provider. Is it intended?

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	// Convert to API types
+	result := make([]resource_manager.ServiceTypeInstance, len(instances))
+	for i, inst := range instances {
+		result[i] = *ModelToAPI(&inst)
+	}
+
+	return &resource_manager.ServiceTypeInstanceList{
+		Instances: &result,
```
gciavarrini: The API supports `page_token` and `next_page_token`, but this method never sets them.

internal/service/resource_manager/service_type_instance_test.go
```diff
@@ -0,0 +1,346 @@
...
+		instanceService = rmsvc.NewInstanceService(dataStore)
+		ctx = context.Background()
+	})
+
+	AfterEach(func() {
+		mockProvider.Close()
+		dataStore.Close()
+	})
+
+	Describe("CreateInstance", func() {
```
gciavarrini, line 80: There is no test for the case for provider exists but HealthStatus != Ready

internal/service/resource_manager/service_type_instance_test.go
```diff
@@ -0,0 +1,346 @@
...
+			}
+
+			_, err := instanceService.CreateInstance(ctx, req, nil)
+
+			Expect(err).To(HaveOccurred())
+			var svcErr *service.ServiceError
+			Expect(err).To(BeAssignableToTypeOf(svcErr))
+			errors.As(err, &svcErr)
+			Expect(svcErr.Code).To(Equal(service.ErrCodeProviderError))
+		})
```
gciavarrini, line 209: Missing test for "provider creates the instance but DB record creation fails, the API should return 500 with instanceId in the error." as described in enhancements [sp-resource-manager.md](https://github.com/dcm-project/enhancements/blob/843b5c4c4f6a68c75773dd5a401db0e72555d59a/enhancements/sp-resource-manager/sp-resource-manager.md?plain=1#L221-L224)

gciavarrini: Also, there is no test for when the provider responds with an HTTP error `4xx` or `5xx`

internal/service/convert.go
```diff
@@ -19,8 +19,8 @@ func ModelToProvider(m *model.Provider) *server.Provider {
 		SchemaVersion: m.SchemaVersion,
```
gciavarrini, line 19: just a reminder: this will be removed

internal/service/resource_manager/service_type_instance_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+			Expect(err).To(HaveOccurred())
+			var svcErr *service.ServiceError
+			Expect(err).To(BeAssignableToTypeOf(svcErr))
+			errors.As(err, &svcErr)
+			Expect(svcErr.Code).To(Equal(service.ErrCodeValidation))
+		})
+	})
+
+	Describe("ListInstances", func() {
```
gciavarrini, line 360: There is no test that the provider filter works

internal/service/resource_manager/convert.go
jenniferubah: We have not defined the output schema that's why it is still unclear. And you're right, in the enhancement it is `instanceId`. Will update this in another PR and enhancement when we have the schema

internal/service/resource_manager/convert.go
jenniferubah: We haven't defined that yet so did not add it. It will be in another PR for sure.

internal/service/resource_manager/service_type_instance_test.go
jenniferubah: Done, added the tests for the HTTP errors and DB creation failure.

internal/service/convert.go
jenniferubah: Done- Removed it in the tests

internal/service/resource_manager/service_type_instance_test.go
jenniferubah: Done, added test for provider filter

internal/service/resource_manager/service_type_instance_test.go
jenniferubah: Done, added test

internal/service/resource_manager/service_type_instance.go
jenniferubah: Good catch, updated it to return 500 error. Thanks!

internal/handlers/resource_manager/convert.go
jenniferubah: Actually, I was missing the Status field here. Added it now. Thanks
The note is mainly for the actual value of the status field as well as the status monitoring implementation

internal/handlers/resource_manager/handler.go
jenniferubah: That's right - I had missed it in a previous PR. I wanted to do this change in another PR but I have added it to this PR. Updated to the param name to provider.

internal/service/resource_manager/service_type_instance.go
jenniferubah: Updated this

internal/handlers/resource_manager/errors.go
jenniferubah: Updated this to 404 and added 500 as well to the list.

internal/service/resource_manager/service_type_instance.go
jenniferubah: Updated the log message

jenniferubah: >BTW, do we want to have a routine checking for discrepancies between SP and internal DB? Maybe it will be part of the status monitoring PR as the SP will be sending updates about its resources periodically.

I believe this would be part of the status monitoring flow and ensure the SP and DB data are in sync. I know right now in the design, it only updating the status field but I understand in future version the output configuration data will be sent( to internal DB) as part of it.  So far, that's my understanding.

jenniferubah:
> just something to take care of in case the failure occurs when interacting with local DB after successful request from SP

That's right. I plan to add the re-try mechanism/error handliang + logging in the next PR as part of the [FLPATH-3207](https://issues.redhat.com/browse/FLPATH-3207)) ticket

internal/service/resource_manager/service_type_instance.go
jenniferubah: Done, updated

internal/service/resource_manager/service_type_instance.go
jenniferubah: You're right, I will add the re-try mechanism for both the create and delete failure  and proper error message in the next PR. Just to keep this PR within scope and not too long

internal/service/resource_manager/service_type_instance.go
gabriel-farache: ok we can go this way and if we see that's not the case we can always implement later

internal/service/resource_manager/service_type_instance.go
gabriel-farache: ok, fine with me

api/v1alpha1/resource_manager/openapi.yaml
```diff
@@ -36,17 +36,17 @@ paths:
...
-  /service-types-instances:
+  /service-type-instances:
     get:
       tags:
         - instance
       summary: List all service type instances
       operationId: listInstances
       description: Returns a list of service type instances with optional filtering
       parameters:
-        - name: type
```
gabriel-farache, line 47: why removing filter on type? A provider may implement several types in theory, did you change that because you consider/assume a 1:1 mapping between provider and types?
If that is the case, we can keep as it is for now and make it evolve later, once the assumption will not be true anymore

internal/handlers/resource_manager/handler.go
gabriel-farache, line 1: why did you name this folder `resource_manager` instead of `service_type`?  The `resource_manager` makes me think it's related to the Placement Manager (https://github.com/dcm-project/placement-manager/blob/main/api/v1alpha1/openapi.yaml#L39)

And why did you not created another folder `provider` that would contain the files for the `providers` endpoint?

gabriel-farache (commented):
Having `service-types` defined in https://github.com/dcm-project/catalog-manager and `service-type-instances` here is good for brain gymnastic. But the names of the folder `resource_manager` holding all `service-type-instances` related files is making things even more confused to my poor brain with the Placement Manager.
Could you rename the folders and maybe create other ones for the `provider` related files so all is sorted properly?

internal/service/resource_manager/service_type_instance.go
jenniferubah: Sounds good

api/v1alpha1/resource_manager/openapi.yaml
jenniferubah: I removed it because we decided not to pass the `serviceType` field and hence not have it in the DB column. However, I see the value of filtering instances via type so we can extract the service type from the provider information and save it a column during creation, then we can add the filtering for this endpoint? is that what you were thinking or what do you think?

internal/handlers/resource_manager/handler.go
jenniferubah: >why did you name this folder resource_manager instead of service_type? The resource_manager makes me think it's related to the Placement Manager (https://github.com/dcm-project/placement-manager/blob/main/api/v1alpha1/openapi.yaml#L39)

I named it `resource_manager` to reflect the name of the service manager and compared to `service_type_instance`, i think it was shorter. Calling the folder `service_type` might be more confusing though or did you mean `service_type_instance`? i can rename it to `service_type_instance` if it's less confusing.

jenniferubah: >And why did you not created another folder provider that would contain the files for the providers endpoint?

 The plan is to do the folder structure refactoring afterwards. I will create a ticket to track this

internal/service/resource_manager/service_type_instance.go
jenniferubah: Done, added as well

internal/handlers/resource_manager/handler.go
gabriel-farache: `service_type_instance` sounds better indeed

> The plan is to do the folder structure refactoring afterwards. I will create a ticket to track this

ok cool

api/v1alpha1/resource_manager/openapi.yaml
gabriel-farache: > I removed it because we decided not to pass the serviceType field and hence not have it in the DB column

Ok, I missed that

it would be great to be able to filter by service type and as you mentioned, we can get that from the provider metadata. 

Not sure it's needed for now, we may want to focus on the essentials things as the summit will be there soon enough

So you can either discard the comment or  create a jira if you think this will be needed

jenniferubah: Sure, I created a ticket for it: [FLPATH-3212](https://issues.redhat.com/browse/FLPATH-3212)

internal/handlers/resource_manager/handler.go
jenniferubah: Created a ticket for this: [FLPATH-3208](https://issues.redhat.com/browse/FLPATH-3208)

internal/service/resource_manager/convert.go
jenniferubah: [FLPATH-3193](https://issues.redhat.com/browse/FLPATH-3193) ticket to track the output schema design/implementation

cmd/service-provider-manager/main.go
```diff
@@ -32,16 +34,21 @@ func main() {
...
+	rmHandler := rmhandlers.NewHandler(instanceService)
 
 	// Start server
 	listener, err := net.Listen("tcp", cfg.Service.Address)
 	if err != nil {
 		log.Fatalf("Failed to listen: %v", err)
 	}
 
-	srv := apiserver.New(cfg, listener, handler)
+	srv := apiserver.New(cfg, listener, providerHandler, rmHandler)
```
ygalblum, line 51: The Provider Service is meant for external components - SPs registering themselves. However, I'm not sure if the Instance Service should be open externally. Who other than the placement manager is supposed to use this API? If it's internal only, I think it should listen on a separate port.
@dcm-project/team-dcm WDYT?

internal/handlers/resource_manager/handler_test.go
```diff
@@ -0,0 +1,321 @@
...
+			Expect(ok).To(BeTrue())
+
+			// Second creation with same ID should fail
+			resp2, err := handler.CreateInstance(ctx, req)
+			Expect(err).NotTo(HaveOccurred())
+			_, ok = resp2.(server.CreateInstance409ApplicationProblemPlusJSONResponse)
+			Expect(ok).To(BeTrue())
+		})
+
+		It("returns 400 for non-existent provider", func() {
```
ygalblum: IIUC if the provider doesn't exist, the service returns 404 and so should the handler. No?

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	// Resolve instance ID
+	instanceID, err := s.resolveInstanceID(ctx, queryID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert spec to JSON
+	specJSON, err := json.Marshal(request.Spec)
```
ygalblum: Why is this needed? Only for validation? If yes, does it need to be here? Because the SP will have to validate it anyhow and will do it better as it knows the actual expected structure, not just that it is a JSON

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,320 @@
...
+	if err != nil {
+		return nil, &service.ServiceError{
+			Code:    service.ErrCodeValidation,
+			Message: fmt.Sprintf("invalid spec: %v", err),
+		}
+	}
+
+	// Send request to provider endpoint
+	providerResponse, err := s.sendToProvider(ctx, provider.Endpoint, request)
```
ygalblum: IIUC, `sendToProvider` does not alter the request. As a result, the ID field is not set. So, the Provider will generate its own ID, but we don't store that ID here. As a result, when trying to delete, we won't have the ID the SP holds.
Am I right here?
If so, we need to either set the ID in the request to the SP, or store it. The problem I see with setting the ID is that UUIDs are not supported as a provided ID.

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+// ListInstances returns instances with optional filtering and pagination
+func (s *InstanceService) ListInstances(ctx context.Context, providerName *string, maxPageSize *int, pageToken string) (*resource_manager.ServiceTypeInstanceList, error) {
+	opts := &rmstore.ServiceTypeInstanceListOptions{
+		ProviderName: providerName,
+	}
+
+	// Apply max page size (default 50, max 100)
+	if maxPageSize != nil && *maxPageSize > 0 && *maxPageSize < 100 {
```
ygalblum: Here `List` disregards out of range values, should it return an error?

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,320 @@
...
+		opts.PageSize = *maxPageSize
+	}
+
+	// Apply page token
+	if pageToken != "" {
```
ygalblum: Consider changing `pageToken` to `*string`. The caller got it as a pointer, dereferenced it and now we take a pointer to it

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,320 @@
...
+	for i, inst := range result.Instances {
+		apiInstances[i] = *ModelToAPI(&inst)
+	}
+
+	apiResult := &resource_manager.ServiceTypeInstanceList{
+		Instances: &apiInstances,
+	}
+
+	// Set next page token if available
+	if result.NextPageToken != "" {
```
ygalblum: Similarly, if `NextPageToken` was a pointer, assignment would be much simpler

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,320 @@
...
+			Message: fmt.Sprintf("failed to retrieve provider: %v", err),
+		}
+	}
+
+	// Send delete request to provider if provider still exists
+	if provider != nil {
+		err = s.sendDeleteToProvider(ctx, provider.Endpoint, instanceID)
+		if err != nil {
+			log.Printf("Error: failed to delete instance (%s) from provider (%s): %v", instanceID, provider.Name, err)
+			if errors.Is(err, rmstore.ErrInstanceNotFound) {
```
ygalblum: I'm not sure why the handling is specific to this error. What if we get a different error? For example, if the provider is restarting. If we continue, we'll delete the instance from the DB and not be able to delete it from the provider.
Furthermore, looking at `sendDeleteToProvider` is always uses `fmt.Errorf`, not a defined type

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,320 @@
...
+			Code:    service.ErrCodeConflict,
+			Message: fmt.Sprintf("instance with ID '%s' already exists", requestedID),
+		}
+	}
+
+	return requestedID, nil
+}
+
+// sendToProvider sends the create request to the provider's endpoint
+func (s *InstanceService) sendToProvider(ctx context.Context, endpoint string, request *resource_manager.ServiceTypeInstance) (*ProviderResponse, error) {
```
ygalblum: Consider renaming this method as it is not a generic `sendToProvider`, it's specific for POST request

internal/service/resource_manager/service_type_instance.go
```diff
@@ -0,0 +1,320 @@
...
+
+	resp, err := s.httpClient.R().
+		SetContext(ctx).
+		SetHeader("Content-Type", "application/json").
+		SetBody(request).
+		SetResult(&providerResp).
+		Post(endpoint)
+
+	if err != nil {
+		return nil, &service.ServiceError{
```
ygalblum: Consider adding a function like:
```golang
func NewProviderError(message string) *ServiceError {
	return &ServiceError{
		Code:    ErrCodeProviderError,
		Message: message,
	}
}
```

internal/handlers/resource_manager/handler_test.go
jenniferubah: Yes the handler returns 404. Fixed the typo in the test

internal/service/resource_manager/service_type_instance.go
jenniferubah: This conversion is for storing it in the DB as the spec column is a json type. when sending the payload to the SP, I am passing as is.

internal/service/resource_manager/service_type_instance.go
jenniferubah: Updated to return an error

internal/service/resource_manager/service_type_instance.go
jenniferubah: Done

internal/service/resource_manager/service_type_instance.go
jenniferubah: Done

internal/service/resource_manager/service_type_instance.go
jenniferubah: You're right. Updated it, it's supposed to be generic and not specific

internal/service/resource_manager/service_type_instance.go
jenniferubah: Done - renamed both functions

internal/service/resource_manager/service_type_instance.go
jenniferubah: I was thinking about this for long (whether to send the Id or have SP generate it), eventually I missed adding it.

>The problem I see with setting the ID is that UUIDs are not supported as a provided ID.

Maybe providers should support UUID? I updated to pass the id as query param to the provider.
Although, I'm not opposed to having the SP generating the ID my most concern is, if the request comes with its own ID, then we would have to pass that down to the SPs, right?

internal/service/resource_manager/service_type_instance.go
jenniferubah: Done, refactored the service errors

cmd/service-provider-manager/main.go
jenniferubah: I agree with it could be separate. But since we will be using API gateway, wouldn't that control the endpoint access and hence no need for separate ports? I might be wrong here though

internal/handlers/resource_manager/handler.go
```diff
@@ -29,9 +29,9 @@ func (h *Handler) GetHealth(ctx context.Context, request server.GetHealthRequest
 
 // ListInstances returns a paginated list of service type instances.
 func (h *Handler) ListInstances(ctx context.Context, request server.ListInstancesRequestObject) (server.ListInstancesResponseObject, error) {
-	var pageToken string
+	var pageToken *string
 	if request.Params.PageToken != nil {
-		pageToken = *request.Params.PageToken
+		pageToken = request.Params.PageToken
 	}
```
ygalblum: You don't need this anymore. Just use `request.Params.PageToken` when calling `ListInstances` (just as you do for the other parameters)

internal/service/resource_manager/service_type_instance.go
ygalblum: Are you sure you need it? Can't you assign `Spec: request.Spec` for the DB?

internal/service/resource_manager/service_type_instance.go
ygalblum: Make `result.NextPageToken` a pointer. This will again will remove the need for the if

internal/service/resource_manager/service_type_instance_test.go
```diff
@@ -0,0 +1,603 @@
...
+				Expect(err).NotTo(HaveOccurred())
+			}
+
+			result, err := instanceService.ListInstances(ctx, nil, nil, nil)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(*result.Instances).To(HaveLen(3))
+		})
+
+		It("respects max page size", func() {
```
ygalblum: There are several tests for pagination. I think you can consolidate them into one.
- Create 5 instances
- Call list with max of 2 check that the length is 2 and the token is not empty
- Call list again with max of 2 and check that the length is 2 and the token is not empty (same as the first)
- Call list again with max of 2. Now, the length is supposed to be 1 and the token should be empty
I think that covers all the cases covered here in many tests

internal/service/resource_manager/service_type_instance.go
jenniferubah: On the DB level, it didn't work since I had the type as `datatypes.JSON`
but it works  when i set it on the gorm level like this
`	Spec         map[string]interface{} `gorm:"type:jsonb;serializer:json"`
I can update the code to use this and remove conversion.

cmd/service-provider-manager/main.go
gabriel-farache: Either the API gateway or the AuthZ would filter the authorized requests at some point so I am not sure we need to worry about that now
If we see that it's not good to have all in the same port, we can easily make it evolve later

On a side note, if we decide to go that way, we need to apply that to all other services.

internal/service/resource_manager/service_type_instance.go
gabriel-farache: if you do that, the insert will fail in case of malformated spec, correct? 
If we want to fully delegate the validation to the SP we may simply store the spec as a string?
Failing fast on this type of format validation may be relevant though

internal/service/resource_manager/service_type_instance_test.go
jenniferubah: Updated it, see [41e6254](https://github.com/dcm-project/service-provider-manager/pull/18/commits/41e625487724e2b3da7ea4409912cbd512b0d513)

internal/handlers/resource_manager/handler.go
jenniferubah: Updated it, see [41e6254](https://github.com/dcm-project/service-provider-manager/pull/18/commits/41e625487724e2b3da7ea4409912cbd512b0d513)

internal/service/resource_manager/service_type_instance.go
jenniferubah: Updated it, see [41e6254](https://github.com/dcm-project/service-provider-manager/pull/18/commits/41e625487724e2b3da7ea4409912cbd512b0d513)

internal/service/resource_manager/service_type_instance.go
jenniferubah: >if you do that, the insert will fail in case of malformated spec, correct?

Yes it would fail if it's not a json spec. The handler would already catch this though before it reaches the DB layer.

>If we want to fully delegate the validation to the SP we may simply store the spec as a string?

We would still have to marshal and unmarshal manually so why not do it at gorm level?

>Failing fast on this type of format validation may be relevant though

Not really a fail fast because the storing at the DB happens after SP validates the spec. The intention here is to preserve the input structure provided by the user. At least that's how i see it.

gabriel-farache: > > If we want to fully delegate the validation to the SP we may simply store the spec as a string?
> 
> We would still have to marshal and unmarshal manually so why not do it at gorm level?

Ah, right, `spec` in an `object` markes as additional properties in the openapi spec so just passing the string from place to place cannot be done

> > Failing fast on this type of format validation may be relevant though
> 
> Not really a fail fast because the storing at the DB happens after SP validates the spec. The intention here is to preserve the input structure provided by the user. At least that's how i see it.

What I meant is that keeping the validation you did here would fail fast without even sending anything to SP. This check should even occurs before evaluating the policies and just after.

On second thought, as the check occurs just before the requests is sent to the SP, I think it's indeed not needed as it would not save anything relevant. And if the SP validates the `spec` there is no reason gorm fails it.

So except if we consider that `CreateInstance` receives a `spec` already validated by the previous layers (I think this API is internal, right? It's called by the Placement Manager once the policies were evaluated), your gorm annotation seems to be a good solution

jenniferubah: >What I meant is that keeping the validation you did here would fail fast without even sending anything to SP

Ah yes, sorry you're right. I had already removed the validation in my local code and my mind was already referring to gorm and DB layer. I'll push the code now.

>I think this API is internal, right? It's called by the Placement Manager once the policies were evaluated

That's correct

jenniferubah: Updated, see here: [b7787e1](https://github.com/dcm-project/service-provider-manager/pull/18/changes/b7787e1222b118ef4eac9abaaa3a91daf57de418)

cmd/service-provider-manager/main.go
ygalblum: > On a side note, if we decide to go that way, we need to apply that to all other services.

That's the plan for the Policy Manager. This is why I defined the policy evaluation separately and plan to expose a separate port for it.

internal/store/model/service_type_instance.go
```diff
@@ -4,17 +4,16 @@ import (
...
-	Status       string         `gorm:"column:status;not null"`
-	InstanceName string         `gorm:"column:instance_name;not null"`
-	Spec         datatypes.JSON `gorm:"column:spec;not null"`
-	CreateTime   time.Time      `gorm:"column:create_time;autoCreateTime"`
-	UpdateTime   time.Time      `gorm:"column:update_time;autoUpdateTime"`
+	ID           uuid.UUID              `gorm:"primaryKey;type:uuid"`
+	ProviderName string                 `gorm:"column:provider_name;not null"`
+	Status       string                 `gorm:"column:status;not null"`
+	InstanceName string                 `gorm:"column:instance_name;not null"`
+	Spec         map[string]interface{} `gorm:"column:spec;type:jsonb;serializer:json;not null"`
```
ygalblum: Some IDEs give a warning here to use `map[string]any` instead of `map[string]interface{}`

internal/service/resource_manager/service_type_instance.go
ygalblum: I added a small comment on the commit

internal/service/resource_manager/service_type_instance.go
ygalblum: > Maybe providers should support UUID? 

While it's not 100% AEP, I'm fine with that.

> Although, I'm not opposed to having the SP generating the ID

if the SP generated the ID, then the SPRM will need to store the value. It's all good, just need to remember

> my most concern is, if the request comes with its own ID, then we would have to pass that down to the SPs, right?

I'm not sure that it's a must. The API guaranties that the ID of the resource created by a POST call will be the one passed. Not the ID of the resource created down the line.

internal/store/model/service_type_instance.go
jenniferubah: hmm, interesting, Intellij doesn't but Cursor does. On the api spec level, it has interface so I'm leaning towards keeping it the same. then again, it's verbose, I'll change it

internal/service/resource_manager/service_type_instance.go
jenniferubah: >if the SP generated the ID, then the SPRM will need to store the value. It's all good, just need to remember

We can go this way then. I can added the provider instance id field to the db layer and api layer to ensure the linking. Created this [ticket](https://issues.redhat.com/browse/FLPATH-3259) to track this and can do it in a follow up PR.

gabriel-farache (approved):

cmd/service-provider-manager/main.go
jenniferubah: Also, added this to the agenda for next week.

ygalblum: 👍🏼

ygalblum (approved):

ygalblum:
Not sure why the required checks are not reporting their status

jenniferubah:
> Not sure why the required checks are not reporting their status


Same, not sure either. I see 5 checks are green but not the others. I'll squash/merge this for now though so it's not blocking.

PR #24, opened 2026-02-25
Add build-push-quay workflow to push images to Quay

gciavarrini:
Adds a workflow that builds the container image and pushes it to quay.io/dcm-project/service-provider-manager.

Triggers:
- Push to main (tags: latest, sha-xxx)
- Push of version tags v* (tags: latest, sha-xxx, v1.0.0)
- Manual run (optional version input)

### DIFF
--- a/.github/workflows/build-push-quay.yaml
+++ b/.github/workflows/build-push-quay.yaml
@@ -1,0 +1,21 @@
+name: Build and Push Image
+
+on:
+  push:
+    branches: [main]
+    tags: ['v*']
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version tag to push (e.g. v1.0.0). When set, only this tag is used.'
+        required: false
+
+jobs:
+  build-push:
+    uses: dcm-project/shared-workflows/.github/workflows/build-push-quay.yaml@main
+    with:
+      image-name: service-provider-manager
+      version: ${{ github.event.inputs.version }}
+    secrets:
+      quay-username: ${{ secrets.QUAY_USERNAME }}
+      quay-password: ${{ secrets.QUAY_PASSWORD }}


.github/workflows/build-push-quay.yaml
```diff
@@ -0,0 +1,39 @@
...
+    branches: [main]
+    tags: ['v*']
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version tag to push (e.g. v1.0.0). Leave empty for latest+sha only.'
+        required: false
+
+jobs:
+  prepare:
```
ygalblum: This seems like something we're gonna repeat. Should it be implemented in the common repo? (I'm assuming this workflow can get the outputs)

.github/workflows/build-push-quay.yaml
```diff
@@ -0,0 +1,39 @@
...
+      tags: ${{ steps.tags.outputs.tags }}
+    steps:
+      - uses: actions/checkout@v6
+      - id: tags
+        run: |
+          TAGS="latest,sha-${GITHUB_SHA:0:7}"
+          if [[ "${{ github.ref }}" == refs/tags/* ]]; then
+            TAGS="${TAGS},${{ github.ref_name }}"
+          fi
+          if [[ -n "${{ github.event.inputs.version }}" ]]; then
```
ygalblum: I think that if `version` is passed (i.e. manual trigger), it should be the only one used

.github/workflows/build-push-quay.yaml
gciavarrini: good point. 
Please check https://github.com/dcm-project/shared-workflows/pull/7 and i'll update this part accordigly

.github/workflows/build-push-quay.yaml
gciavarrini: See https://github.com/dcm-project/shared-workflows/pull/7

ygalblum (approved):

gabriel-farache (approved):

PR #26, opened 2026-03-03
Allow passing UUID as id in create operation

ygalblum:
Broaden the regex of the id parameter to allow for UUID


gabriel-farache (approved):
Is there any update needed on the enhancement doc that mentioned the limitation we are removing?

ygalblum:
> Is there any update needed on the enhancement doc that mentioned the limitation we are removing?

I don't see any reference to it here: https://github.com/dcm-project/enhancements/blob/1f5aa72711c05be79a6e255acfd57de498dad75f/enhancements/sp-resource-manager/sp-resource-manager.md

gciavarrini (approved):

machacekondra (approved):

PR #30, opened 2026-03-10
Convert ID format from uuid to string

jenniferubah:
Change ID format from UUID to string in openapi, handlers, service, store and tests.

Assisted-By: Cursor (Claude 4.5 Sonnet)


internal/handlers/resource_manager/handler_test.go
```diff
@@ -191,18 +191,6 @@ var _ = Describe("Resource Manager Handler", func() {
 			_, ok := resp.(server.GetInstance404ApplicationProblemPlusJSONResponse)
 			Expect(ok).To(BeTrue())
 		})
-
-		It("returns 400 for invalid ID format", func() {
```
gabriel-farache, line 195: about the format of the ID, is there a middleware validating it in the request before the request reaches the handler?

api/v1alpha1/openapi.yaml
ygalblum: Move this to a common schema (to be used by all the operation in this endpoint (like in https://github.com/dcm-project/policy-manager/blob/15bd45617e873e70879e67c12262ff273ebb4f63/api/v1alpha1/openapi.yaml#L307)

jenniferubah: Done, [d9b45a8](https://github.com/dcm-project/service-provider-manager/pull/30/commits/d9b45a8fe07d11584ca3c3cec5323dffc69f9559)

ygalblum (approved):

internal/handlers/resource_manager/handler_test.go
jenniferubah: Thanks, this was missing. Added the middleware validation for each api spec: [43aea4c](https://github.com/dcm-project/service-provider-manager/pull/30/commits/43aea4c208fc8d8d524917193890d9f450ed40d3)

gabriel-farache (approved):

PR #31, opened 2026-03-11
Retry mechanism with exponential backoff

jenniferubah:
Added retry for creating and deleting instances in DB.

Assisted-By: Cursor

### DIFF
--- a/internal/service/resource_manager/service_type_instance.go
+++ b/internal/service/resource_manager/service_type_instance.go
@@ -162,7 +162,7 @@ func (s *InstanceService) DeleteInstance(ctx context.Context, instanceID string)
 		log.Printf("Deleted instance (%s) from SP (%s)", instanceID, provider.Name)
 	}
 
-	// Delete from database with retry mechanism
+	// Delete from database
 	err = s.store.ServiceTypeInstance().Delete(ctx, instanceID)
 	if err != nil {
 		return service.NewInternalError(fmt.Sprintf("failed to delete database record for instance %s: %v", instanceID, err))
--- a/internal/service/resource_manager/service_type_instance.go
+++ b/internal/service/resource_manager/service_type_instance.go
@@ -25,7 +25,8 @@ func NewInstanceService(store store.Store) *InstanceService {
 	client := resty.New().
 		SetTimeout(30 * time.Second).
 		SetRetryCount(3).
-		SetRetryWaitTime(1 * time.Second)
+		SetRetryWaitTime(2 * time.Second).
+		SetRetryMaxWaitTime(30 * time.Second)
 
 	return &InstanceService{
 		store:      store,
@@ -73,7 +74,6 @@ func (s *InstanceService) CreateInstance(ctx context.Context, request *resource_
 
 	created, err := s.store.ServiceTypeInstance().Create(ctx, instance)
 	if err != nil {
-		// add re-try mechanism
 		return nil, service.NewInternalError(fmt.Sprintf("failed to create database record for instance %s: %v", providerResponse.ID, err))
 	}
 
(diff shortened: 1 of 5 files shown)


ygalblum (commented):
I wonder if the service is the right place to do the retry and not the store. The store is the one interacting with the external service (the DB), not the service.
Also, have you considered using an external package for this? E.g. https://github.com/cenkalti/backoff/tree/v5

jenniferubah:
> I wonder if the service is the right place to do the retry and not the store. The store is the one interacting with the external service (the DB), not the service. Also, have you considered using an external package for this? E.g. https://github.com/cenkalti/backoff/tree/v5

You're right, having it in the store is a better approach. Also checked out the retry package and this github.com/hashicorp/go-retryablehttp. cenkalti seems the most popular so went it. Thanks!

Refactored the implementation

internal/service/resource_manager/service_type_instance.go
```diff
@@ -162,10 +162,9 @@ func (s *InstanceService) DeleteInstance(ctx context.Context, instanceID string)
 		log.Printf("Deleted instance (%s) from SP (%s)", instanceID, provider.Name)
 	}
 
-	// Delete from database
+	// Delete from database with retry mechanism
```
ygalblum: I think this change should be reverted as well

internal/store/resource_manager/service_instance.go
```diff
@@ -100,21 +102,54 @@ func (s *ServiceTypeInstanceStore) List(ctx context.Context, opts *ServiceTypeIn
 }
 
 func (s *ServiceTypeInstanceStore) Create(ctx context.Context, instance model.ServiceTypeInstance) (*model.ServiceTypeInstance, error) {
-	if err := s.db.WithContext(ctx).Clauses(clause.Returning{}).Create(&instance).Error; err != nil {
+	// Configure exponential backoff retry strategy
+	b := backoff.NewExponentialBackOff()
```
ygalblum: Consider moving the initialization of `backoffWithRetries` to a helper function since the code repeats itself

internal/store/resource_manager/service_instance.go
```diff
@@ -5,7 +5,9 @@ import (
 	"encoding/base64"
 	"errors"
 	"strconv"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
```
ygalblum: From what I can tell the current version is `v5`. See their example: https://github.com/cenkalti/backoff/blob/b67cefebfef38f2da908ed5f60db19dd60c0c8ce/example_test.go#L11

internal/service/resource_manager/service_type_instance.go
jenniferubah: Reverted [d8f4e6b](https://github.com/dcm-project/service-provider-manager/pull/31/commits/d8f4e6bcbca6aad8e919123818ddecbc52bc59ea)

internal/store/resource_manager/service_instance.go
jenniferubah: Refactored [d8f4e6b](https://github.com/dcm-project/service-provider-manager/pull/31/commits/d8f4e6bcbca6aad8e919123818ddecbc52bc59ea)

internal/store/resource_manager/service_instance.go
jenniferubah: Updated to v5 -[d8f4e6b](https://github.com/dcm-project/service-provider-manager/pull/31/commits/d8f4e6bcbca6aad8e919123818ddecbc52bc59ea)

ygalblum (approved):

gabriel-farache (approved):

PR #32, opened 2026-03-16
Service type instance e2e test

jenniferubah:
- Added end to end test for service type instance
- Update make target to always clean up after e2e test

### DIFF
--- a/test/e2e/compose.yaml
+++ b/test/e2e/compose.yaml
@@ -1,14 +1,14 @@
 services:
   postgres:
-    image: docker.io/library/postgres:16-alpine
+    image: quay.io/sclorg/postgresql-16-c9s:latest
     environment:
-      POSTGRES_DB: service-provider
-      POSTGRES_USER: admin
-      POSTGRES_PASSWORD: adminpass
+      POSTGRESQL_DATABASE: service-provider
+      POSTGRESQL_USER: admin
+      POSTGRESQL_PASSWORD: adminpass
     ports:
       - "5432:5432"
     volumes:
-      - postgres_data:/var/lib/postgresql/data
+      - postgres_data:/var/lib/pgsql/data
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -U admin -d service-provider"]
       interval: 5s
@@ -16,7 +16,7 @@ services:
       retries: 5
 
   provider-wiremock:
-    image: docker.io/wiremock/wiremock:3.12.1
+    image: wiremock/wiremock:3x
     ports:
       - "9090:8080"
     healthcheck:
--- a/test/e2e/compose.yaml
+++ b/test/e2e/compose.yaml
@@ -1,0 +1,54 @@
+services:
+  postgres:
+    image: docker.io/library/postgres:16-alpine
+    environment:
+      POSTGRES_DB: service-provider
+      POSTGRES_USER: admin
+      POSTGRES_PASSWORD: adminpass
+    ports:
+      - "5432:5432"
+    volumes:
+      - postgres_data:/var/lib/postgresql/data
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U admin -d service-provider"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+  provider-wiremock:
+    image: docker.io/wiremock/wiremock:3.12.1
+    ports:
+      - "9090:8080"
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8080/__admin/health"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+  service-provider-manager:
+    build:
+      context: ../../
+      dockerfile: Containerfile
+    environment:
+      DB_TYPE: pgsql
+      DB_HOST: postgres
+      DB_PORT: "5432"
+      DB_NAME: service-provider
+      DB_USER: admin
+      DB_PASS: adminpass
+      SVC_ADDRESS: ":8080"
+    ports:
+      - "8080:8080"
+    depends_on:
+      postgres:
+        condition: service_healthy
+      provider-wiremock:
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8080/api/v1alpha1/health"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+volumes:
+  postgres_data:
--- a/test/e2e/service_instance_test.go
+++ b/test/e2e/service_instance_test.go
@@ -18,14 +18,15 @@ import (
 
 var _ = Describe("Service Instance API", func() {
 	var (
-		rmApiClient *rmClient.ClientWithResponses
-		apiClient   *client.ClientWithResponses
-		ctx         context.Context
-		baseURL     string
+		rmApiClient  *rmClient.ClientWithResponses
+		apiClient    *client.ClientWithResponses
+		ctx          context.Context
+		providerID   string
+		providerName string
 	)
 
 	BeforeEach(func() {
-		baseURL = os.Getenv("API_URL")
+		baseURL := os.Getenv("API_URL")
 		if baseURL == "" {
 			baseURL = "http://localhost:8080/api/v1alpha1"
 		}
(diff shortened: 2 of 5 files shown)


test/e2e/service_instance_test.go
```diff
@@ -0,0 +1,281 @@
...
+			customID := uuid.New().String()
+			params := &resource_manager.CreateInstanceParams{Id: &customID}
+
+			createResp, err := rmApiClient.CreateInstanceWithResponse(ctx, params, resource_manager.ServiceTypeInstance{
+				ProviderName: providerName,
+				Spec:         map[string]interface{}{"cpu": 2, "memory": "4GB"},
+			})
+
+			Expect(err).NotTo(HaveOccurred())
+			// Will fail because example.com is not a real provider, but ID should be validated
```
ygalblum: Not sure I understand this comment

test/e2e/service_instance_test.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+
+			listResp, err := rmApiClient.ListInstancesWithResponse(ctx, params)
+			Expect(err).NotTo(HaveOccurred())
+			// Invalid page token is treated as empty, returns 200 with results from start
+			Expect(listResp.StatusCode()).To(Equal(http.StatusOK))
+		})
+	})
+
+	Describe("Service Instance Get Operation", func() {
```
ygalblum: A successful case is missing

test/e2e/service_instance_test.go
```diff
@@ -0,0 +1,281 @@
...
+			nonExistentID := uuid.New().String()
+			getResp, err := rmApiClient.GetInstanceWithResponse(ctx, nonExistentID)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(getResp.StatusCode()).To(Equal(http.StatusNotFound))
+		})
+	})
+
+	Describe("Service Instance Delete Operation", func() {
+		It("returns 404 when deleting non-existent instance", func() {
```
ygalblum, line 287: A successful case is missing

machacekondra (approved):

test/e2e/service_instance_test.go
jenniferubah: Added now - Used wiremock to mock a provider and refactored the test structure as well.
[28ade1b](https://github.com/dcm-project/service-provider-manager/pull/32/commits/28ade1b86d85c985420f9b433623e834345b8f03)

test/e2e/service_instance_test.go
jenniferubah: Added now - Used simple httptest to mock a provider and refactored the test structure as well.
[53d6e53](https://github.com/dcm-project/service-provider-manager/pull/32/commits/53d6e530db098382900d2c2056903763c1bbb503)

jenniferubah: Actually, it's not working on the ci, checking

test/e2e/service_instance_test.go
jenniferubah: AI shenanigans. Removed the assertion completely. I think it was trying to validate the id even if the failed to send it to the provider.

test/e2e/service_instance_test.go
jenniferubah: Updated now - Used wiremock to mock a provider and refactored the test structure as well.
[28ade1b](https://github.com/dcm-project/service-provider-manager/pull/32/commits/28ade1b86d85c985420f9b433623e834345b8f03)

test/e2e/service_instance_test.go
jenniferubah: I also swapped the e2e to point to the sharedworkflow black-box test and i think that's why the previous e2e in the GitHub Actions does not report a status.

test/e2e/compose.yaml
```diff
@@ -0,0 +1,54 @@
+services:
+  postgres:
+    image: docker.io/library/postgres:16-alpine
```
ygalblum: This is not the image you download in the CI workflow. The change should be here as it aligns with the rest of the repos.
Note that it will also require changes in the environment variables. Take a look at https://github.com/dcm-project/catalog-manager/blob/566e537054933ef15f02634641c59192a3815c33/test/subsystem/docker-compose.yaml#L3

ygalblum:
@jenniferubah I've remove the e2e from the required tests list to allow merging this one. Once merged, we can update the list again (to use the one from the blackbox testing)

test/e2e/compose.yaml
jenniferubah: Thanks. Updated the image and env variable. [efd2a19](https://github.com/dcm-project/service-provider-manager/pull/32/commits/efd2a194db2c2302dd00371e830b3ce184da1b1e)

jenniferubah:
> @jenniferubah I've remove the e2e from the required tests list to allow merging this one. Once merged, we can update the list again (to use the one from the blackbox testing)

Sure, will update it once merged. Thank you

ygalblum (approved):

test/e2e/service_instance_test.go
```diff
@@ -0,0 +1,295 @@
...
+			nonExistentID := uuid.New().String()
+			getResp, err := rmApiClient.GetInstanceWithResponse(ctx, nonExistentID)
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(getResp.StatusCode()).To(Equal(http.StatusNotFound))
+		})
+	})
+
+	Describe("List Instances", func() {
+		It("returns created instances in the list", func() {
```
gabriel-farache, line 190: do we want to have a test for empty?

test/e2e/service_instance_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+			Expect(resp.JSON200).NotTo(BeNil())
+			Expect(*resp.JSON200.Status).To(Equal("ok"))
+		})
+	})
+
+	Describe("Create Instance", func() {
+		It("creates an instance with specified ID", func() {
```
gabriel-farache, line 80: do we want a test with malformater user specified ID?

gabriel-farache (commented):
in this e2e test we are only testing DMC => SP but if I remember correctly, the SPRM is also the service managing the registration of SP, do we want to have E2E tests on this as  here:
1. send POST with the same info as what a SP will provide (the endpoints, the service type served, ...) on the registration endpoint
2. use the registered SP to make the requests. This would be triggered by a REST call on the SPRM endpoint to create/list/get/delete the resource 


Because here you are using the client directly, the step which get the SP endpoint and other information is skipped

Here it looks like more integration tests than E2E but I may be missing something

jenniferubah:
> SPRM is also the service managing the registration of SP, do we want to have E2E tests on this as here:

The Provider API handles the registration of the SPs, while the SPRM API handles the creation of the instances. And during the creation of the instances, SPRM queries the db to get the provider's information, including the endpoint, and then send it to the provider's endpoint. Hence in this e2e, we are not testing the registration flow because the provider should already exist in the database.
Also, e2e for the provider registration is already covered here: https://github.com/dcm-project/service-provider-manager/blob/main/test/e2e/provider_test.go

>Here it looks like more integration tests than E2E

You're right, this is mostly an integration test. A full E2E would be what you described. The idea was to test just the SPRM API and since the test/e2e folder exists, it made sense to put it in there. However, we can add a full E2E test in the future PR and rename these ones as subsystem/integration tests. What do you think?

test/e2e/service_instance_test.go
jenniferubah: Would this be caught by the middleware validator before it reaches the handler? or am I misunderstanding

test/e2e/service_instance_test.go
jenniferubah: I'm not sure if there is any need for it since it is already covered in the unit test and here, it would require deleting the instances created in `BeforeEach`. Since the unit test will catch it, I'm thinking there is less value in it.

test/e2e/service_instance_test.go
gabriel-farache: ah, yes you are right, this is catch before it reaches the service

test/e2e/service_instance_test.go
gabriel-farache: the idea is to make sure that the services are behaving well with 0 row in the DB
If there are lower level tests ensuring that when the mock DB returns 0 then the services behaves as expected, I agree that is OK not to have it here (we will need it in the real E2E tests with call to REST endpoints)

gabriel-farache (approved):

gabriel-farache:
> Also, e2e for the provider registration is already covered here: https://github.com/dcm-project/service-provider-manager/blob/main/test/e2e/provider_test.go

Ah, great!


> You're right, this is mostly an integration test. A full E2E would be what you described. The idea was to test just the SPRM API and since the test/e2e folder exists, it made sense to put it in there. However, we can add a full E2E test in the future PR and rename these ones as subsystem/integration tests. What do you think?

fines with me, we are aligned :)

test/e2e/service_instance_test.go
jenniferubah: Sure, it makes sense. I can add it in the main E2E.

PR #33, opened 2026-03-17
Resource status reporting implementation

machacekondra:


internal/consumer/consumer.go
```diff
@@ -0,0 +1,177 @@
...
+func (c *StatusConsumer) Stop() {
+	if c.consumeCtx != nil {
+		c.consumeCtx.Stop()
+	}
+	c.conn.Close()
+	log.Println("StatusConsumer stopped")
+}
+
+func (c *StatusConsumer) handleMessage(ctx context.Context, msg jetstream.Msg) {
+	_, err := parseSubject(msg.Subject())
```
ygalblum: We're not using the subject, why bother parsing it?

ygalblum:
The E2E tests are failing because you need to add the NATS service to the compose file: https://github.com/dcm-project/service-provider-manager/blob/main/compose.yaml

internal/store/resource_manager/service_instance.go
```diff
@@ -140,6 +141,23 @@ func (s *ServiceTypeInstanceStore) Get(ctx context.Context, id string) (*model.S
...
 }
 
+func (s *ServiceTypeInstanceStore) UpdateStatus(ctx context.Context, instanceID string, status string, statusMessage string) error {
+	result := s.db.WithContext(ctx).
+		Model(&model.ServiceTypeInstance{}).
+		Where("id = ?", instanceID).
+		Updates(map[string]any{
+			"status":         status,
+			"status_message": statusMessage,
+		})
```
jenniferubah: nit: Couldn't we use the model struct instead of map[string]any?
```
		Updates(model.ServiceTypeInstance{
			Status:        status,
			StatusMessage: statusMessage,
		})
```

internal/consumer/consumer.go
machacekondra: Removing, in future, subject reflect the type of status update (cluster,container,vm) now all are same.

gabriel-farache: > Removing, in future, subject reflect the type of status update (cluster,container,vm) now all are same.

What do you mean by "type of status update"? AFAIK, the subject holds the service type (which is consistent with the examples you give after) 

What do you mean that now they are all the same? We will have different SP so VM, container, Cluster,... so we may want to ensure that the ID and the service type matches no? But I agree we can skip that for now

internal/consumer/consumer.go
```diff
@@ -0,0 +1,159 @@
...
+		o(sc)
+	}
+	return sc, nil
+}
+
+// Option configures a StatusConsumer.
+type Option func(*StatusConsumer)
+
+// WithStreamName sets the JetStream stream name.
+func WithStreamName(name string) Option {
```
gabriel-farache: @ygalblum comment on one of my previous PR https://github.com/dcm-project/k8s-container-service-provider/pull/6#discussion_r2859814088 is relevant on this

just nitpick about the name convention

gabriel-farache (approved):
just a nitpick about the naming convention with `WithXXX` method

internal/consumer/consumer.go
machacekondra: I mean that the structure of CloudEvent is the same for all providers now, it's:
```
Status {
 Id string
 Status string
 Message string
}
```

In future, we may provide different values for different service types, so we would need different format. For example:
```
VmStatus {
 Id string
 Phase string
 DiskStatus string
}

ClusterStatus{
Id string
NodeStatuses {
  Status string
}
}
```

compose.yaml
```diff
@@ -15,6 +15,15 @@ services:
...
       retries: 5
 
+  nats:
+    image: docker.io/library/nats:2-alpine
+    command: ["--jetstream", "--store_dir=/data"]
+    ports:
+      - "4222:4222"
+      - "8222:8222"
+    volumes:
+      - nats_data:/data
```
jenniferubah, line 25: For now, I think we'll need to add this nats deployment to the [e2e compose file](https://github.com/dcm-project/service-provider-manager/blob/main/test/e2e/compose.yaml) go get the e2e ci working. Will clean it up to remove the duplicates between the root compose file and the e2e compose file in another PR after merging this one.

jenniferubah (approved):

internal/consumer/consumer_test.go
```diff
@@ -0,0 +1,180 @@
...
+		cancel()
+		sqlDB, _ := db.DB()
+		_ = sqlDB.Close()
+	})
+
+	publishStatusEvent := func(providerName, serviceType, instanceID, status, message string) {
+		event := cloudevents.New()
+		event.SetID(uuid.New().String())
+		event.SetSource(fmt.Sprintf("dcm/providers/%s", providerName))
+		event.SetType(fmt.Sprintf("dcm.status.%s", serviceType))
```
gabriel-farache, line 86: In the enhancement PR, there `.changed` after: https://github.com/dcm-project/enhancements/pull/37/changes#diff-03b0cbf0bf3764d8d7504f8c55ccf1f3b5b3adaeab3e75361c8257dc9892ebc1R167
Not sure it really matter as wu subscribe to dcm.* but it may be nice to be aligned

ygalblum (approved):

PR #43, opened 2026-03-27
Implement deferred deletion for rehydration flow

ygalblum:
Add soft-delete support to ServiceTypeInstance for the rehydration flow's Delete Old Resource step (Step 5). When the Service Provider is unavailable during deletion, the resource is marked for background cleanup instead of blocking the flow.

Changes:

- Add soft-delete columns to ServiceTypeInstance model: deletion_status (PENDING/FAILED), retry_count, last_deletion_attempt, and deletion_requested_at
- Add DELETE ?deferred=true parameter: on SP failure, marks instance PENDING and returns 204 instead of returning an error
- Add show_deleted query parameter on LIST and GET endpoints, aligned to AEP-164: soft-deleted instances are hidden by default
- Add deletion_status read-only field to ServiceTypeInstance response
- Add background cleanup scheduler (internal/cleanup) that periodically retries PENDING deletions via InstanceService.DeleteFromProvider, marks FAILED after max retries
- Add CleanupConfig (CLEANUP_INTERVAL, CLEANUP_MAX_RETRIES, CLEANUP_TIMEOUT) to application config
- Wire cleanup scheduler into main.go
- Add service-layer and store-layer tests for all soft-delete methods, deferred deletion behavior, and show_deleted filtering


.ai/checkpoints/deferred-deletion.md
```diff
@@ -0,0 +1,112 @@
+# Deferred Deletion for Rehydration Flow
+
+**Commit:** 01f743c
+**Date:** 2026-03-27
+**Context:** [Rehydration Flow Enhancement](https://github.com/dcm-project/enhancements/blob/1f357c1213ccfbb8638f9b5baed82ada86114c15/enhancements/rehydration-flow/rehydration-flow.md), Step 5 — Delete Old Resource
```
jenniferubah: Do we need the `Step 5 ....` here?

internal/cleanup/scheduler.go
```diff
@@ -0,0 +1,107 @@
...
+			s.ProcessPendingDeletions(ctx)
+		}
+	}
+}
+
+// ProcessPendingDeletions attempts to complete all deferred deletions.
+func (s *Scheduler) ProcessPendingDeletions(ctx context.Context) {
+	pending, err := s.store.ServiceTypeInstance().ListPendingDeletions(ctx)
+	if err != nil {
+		log.Printf("Cleanup: error listing pending deletions: %v", err)
```
jenniferubah: We could use slog right and be consistent across the codebase.

.ai/checkpoints/deferred-deletion.md
ygalblum: Right, removed the specific reference

internal/cleanup/scheduler.go
ygalblum: Good catch. I rebased on top of the logging change earlier and told myself to adjust this file and forgot. Thanks

ygalblum:
BTW I had to squash and force push because something in the commit messages was failing the clean commits check

jenniferubah (approved):
I am wondering if we could also have soft-delete for the normal deletion flow as well. I don't envision any disadvantage/concern for it. What do you think? Do you anticipate any concerns?
Note: I know this is out of scope for this PR and the summit so mainly asking in general.

ygalblum:
> I am wondering if we could also have soft-delete for the normal deletion flow as well. I don't envision any disadvantage/concern for it. What do you think? Do you anticipate any concerns? Note: I know this is out of scope for this PR and the summit so mainly asking in general.

I wasn't sure what we'd want to do. Initially I also thought about having all deletes deferred. But, then I thought that maybe for V1 we want "real" delete operations to be sync (that is a successful return of a request to delete guaranties the deletion) 
For V2, we might shift the entire implementation to be async and then all operations will be async.

jenniferubah:
> > I am wondering if we could also have soft-delete for the normal deletion flow as well. I don't envision any disadvantage/concern for it. What do you think? Do you anticipate any concerns? Note: I know this is out of scope for this PR and the summit so mainly asking in general.
> 
> I wasn't sure what we'd want to do. Initially I also thought about having all deletes deferred. But, then I thought that maybe for V1 we want "real" delete operations to be sync (that is a successful return of a request to delete guaranties the deletion) For V2, we might shift the entire implementation to be async and then all operations will be async.

Make sense to me, thanks!

gabriel-farache (approved):

PR #47, opened 2026-04-08
Speed up tests: inject resty client and configurable DB retry for tests

gciavarrini:
## Why
Tests were spending a long time on failures that hit **HTTP retries** (resty) and **DB retries** (backoff around instance `Create` / `HardDelete`). 

Production defaults are right for real outages; in tests they mostly add wall-clock delay without adding signal.

## What
- Optional **`resty.Client`** on the instance service (`nil` → same behavior as before in production).
- Optional **retry options** on the service-type-instance store and **`NewStore`** helper for tests.
- Small **`internal/testutil`** helper for fast backoff in tests so suites stay quick.

### DIFF
--- a/cmd/service-provider-manager/main.go
+++ b/cmd/service-provider-manager/main.go
@@ -61,7 +61,7 @@ func run() int {
 	providerHandler := providerhandler.NewHandler(providerService)
 
 	// Resource Manager API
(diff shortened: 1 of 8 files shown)


jenniferubah (approved):
Thanks!

ygalblum (approved):

PR #46, opened 2026-04-08
FLPATH-3844: Fix drop of optional provider registration data

gciavarrini:
Store optional registration fields in JSON columns and map them on GET/list.
Clear nullable fields on update via explicit column select.

PR #48, opened 2026-04-08
FLPATH-3844: Fix drop of optional provider registration data

gciavarrini:
Store optional registration fields in JSON columns and map them on GET/list.
Clear nullable fields on update via explicit column select.

Replace PR #46

### DIFF
--- a/internal/service/provider/convert.go
+++ b/internal/service/provider/convert.go
@@ -1,6 +1,8 @@
 package provider
 
 import (
+	"encoding/json"
+
 	providerserver "github.com/dcm-project/service-provider-manager/internal/api/server/provider"
 	"github.com/dcm-project/service-provider-manager/internal/service"
 	"github.com/dcm-project/service-provider-manager/internal/store/model"
@@ -8,16 +10,33 @@ import (
 
 // ModelToProvider converts a database model to an API response type.
 func ModelToProvider(m *model.Provider) *providerserver.Provider {
-	return &providerserver.Provider{
+	p := &providerserver.Provider{
 		Id:            &m.ID,
 		Name:          m.Name,
 		ServiceType:   m.ServiceType,
 		SchemaVersion: m.SchemaVersion,
 		Endpoint:      m.Endpoint,
+		DisplayName:   m.DisplayName,
 		HealthStatus:  m.HealthStatus.StringPtr(),
 		CreateTime:    service.PtrTime(m.CreateTime),
 		UpdateTime:    service.PtrTime(m.UpdateTime),
 	}
+	if m.Metadata != nil {
+		b, err := json.Marshal(m.Metadata)
+		if err == nil {
+			var meta providerserver.ProviderMetadata
+			if err := json.Unmarshal(b, &meta); err == nil {
+				p.Metadata = &meta
+			}
+		}
+	}
+	if opBytes, err := json.Marshal(m.Operations); err == nil && string(opBytes) != "null" {
+		var ops []string
+		if err := json.Unmarshal(opBytes, &ops); err == nil {
+			p.Operations = &ops
+		}
+	}
+	return p
 }
 
 // ModelToProviderWithStatus converts a database model to an API response with status.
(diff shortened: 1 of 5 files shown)


internal/service/provider/convert.go
```diff
@@ -29,11 +48,51 @@ func ModelToProviderWithStatus(m *model.Provider, status providerserver.Provider
...
+	b, err := json.Marshal(meta)
+	if err != nil {
+		return nil, err
+	}
+	var m map[string]interface{}
+	if err := json.Unmarshal(b, &m); err != nil {
+		return nil, err
 	}
+	return m, nil
 }
```
jenniferubah, line 98: With the serializer on the gorm level, is this stil needed?

jenniferubah (approved):

internal/service/provider/convert.go
```diff
@@ -29,11 +48,51 @@ func ModelToProviderWithStatus(m *model.Provider, status providerserver.Provider
...
+	return m
+}
+
+func applyProviderRequestToModel(dest *model.Provider, req *providerserver.Provider) {
+	dest.Name = req.Name
+	dest.ServiceType = req.ServiceType
+	dest.SchemaVersion = req.SchemaVersion
+	dest.Endpoint = req.Endpoint
+	dest.DisplayName = req.DisplayName
+	dest.Metadata = nil
```
ygalblum, line 76: What if `dest.Metadata` was not nil and the `req` did not send anything. Should it be set to `nil` or kept?

machacekondra (approved):

internal/service/provider/convert.go
gabriel-farache: With the piece of code that is just after
```
	if req.Metadata != nil {
		if metaMap, err := providerMetadataToMap(req.Metadata); err == nil {
			dest.Metadata = metaMap
		}
	}
```
I guess this line should be deleted no?

gabriel-farache (approved):
1 small comment not blocking

internal/service/provider/convert.go
ygalblum: Yeah, that was my comment.
If `dest.Metadata` is not `nil`, but `req.Metadata` is, then the returned struct will have `Metadata` as `nil`. This means that callers must pass the `Metadata` even when not setting it. Is that the intention?

internal/service/provider/convert.go
gciavarrini: gorm’s `serializer:json` only handles the DB column vs `map[string]interface{}`. 
It doesn’t convert the OpenAPI `ProviderMetadata` struct to that map, so `providerMetadataToMap` (and the reverse in `ModelToProvider`) still does that API <-> map step.

Does it make sense?

internal/service/provider/convert.go
gciavarrini: I keep `dest.Metadata = nil` so that when a field is omitted in the body, stored data is cleared. Without that assignment, omit would keep the previous DB value ("merge" behavior).

Current (omit):
pro: simple “request body is the new state”
cons: clients must resend metadata/operations to keep them.

Alternative (preserve): 
pro: nicer for partial updates; 
cons: clearing can’t be “omit”, we need an explicit rule (null, empty object, etc.) and the same rule for `dest.Operations`.

What do you prefer? @gabriel-farache @ygalblum

internal/service/provider/convert.go
jenniferubah: Yes it makes sense. I missed the part where `ProviderMetadata` is not a typed struct on the DB model layer.

internal/service/provider/convert.go
ygalblum: I think the alternative is better. But, I'm not sure it's worth the hassle for now. I just want to make sure we are intentional and aware that update without metadata will clear it.

ygalblum (approved):

internal/service/provider/convert.go
gciavarrini: @gabriel-farache @ygalblum merging as is. We can revise the decision at any time

PR #51, opened 2026-04-09
Skip provider call for deferred deletion

ygalblum:
## What
Deferred delete now immediately enqueues the instance for background cleanup without attempting to contact the service provider.

## Why
Deferred deletion exists for cases where the provider may be unavailable. The previous implementation tried the provider first, which could block on a timeout before falling back to the cleanup queue.

## Changes
- Move the deferred check before the provider call in DeleteInstance, so deferred=true skips DeleteFromProvider entirely and marks the instance as PENDING immediately
- Simplify the non-deferred error path now that it no longer shares code with the deferred path
- Update tests to assert the provider is not contacted when deferred=true
- Remove unnecessary error-provider mock setup from deferred deletion tests

## Testing
Updated service-layer unit tests to verify the provider delete endpoint is not called in deferred mode, including cases with available providers, missing providers, and FAILED instances.

### DIFF
--- a/internal/service/resource_manager/service_type_instance.go
+++ b/internal/service/resource_manager/service_type_instance.go
@@ -200,7 +200,7 @@ func (s *InstanceService) DeleteInstance(ctx context.Context, instanceID string,
 			}
 		}
 
-		log.Info("Deferred deletion of instance from provider", "instance_id", instance.ID, "provider_name", instance.ProviderName)
+		log.Info("Scheduled deferred deletion of instance from provider", "instance_id", instance.ID, "provider_name", instance.ProviderName)
 		return nil
 	}
 
--- a/internal/service/resource_manager/service_type_instance.go
+++ b/internal/service/resource_manager/service_type_instance.go
@@ -170,8 +170,8 @@ func (s *InstanceService) ListInstances(ctx context.Context, providerName *strin
 	return apiResult, nil
 }
 
-// DeleteInstance removes an instance by ID. When deferred is true, deletion
-// failures are recorded in a cleanup queue and the method returns success.
+// DeleteInstance removes an instance by ID. When deferred is true, the instance
+// is marked for background cleanup without contacting the provider.
 func (s *InstanceService) DeleteInstance(ctx context.Context, instanceID string, deferred bool) error {
 	log := logging.FromContext(ctx)
 	log.Debug("Deleting instance", "instance_id", instanceID, "deferred", deferred)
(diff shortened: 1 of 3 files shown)


jenniferubah (approved):

internal/service/resource_manager/service_type_instance_test.go
```diff
@@ -506,19 +506,17 @@ var _ = Describe("InstanceService", func() {
...
 				ProviderName: "test-provider",
 				Spec:         map[string]interface{}{"cpu": 2},
 			}
 			created, _ := instanceService.CreateInstance(ctx, req, nil)
 
-			// Delete the provider from the database
-			Expect(db.Delete(&model.Provider{}, "name = ?", "test-provider").Error).NotTo(HaveOccurred())
-
-			// Deferred delete should succeed
+			// Deferred delete should succeed without calling the provider
```
gabriel-farache, line 516: how do you validate that the provider was not called, I do not see any `expect` related to this

internal/service/resource_manager/service_type_instance_test.go
```diff
@@ -532,46 +530,26 @@ var _ = Describe("InstanceService", func() {
...
+			// Deferred delete should succeed without attempting provider call
+			err := instanceService.DeleteInstance(ctx, *created.Id, true)
 			Expect(err).NotTo(HaveOccurred())
+			Expect(deleteRequested).To(BeFalse())
 
 			// Verify marked as PENDING
 			result, err := instanceService.ListInstances(ctx, nil, true, nil, nil)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(*result.Instances).To(HaveLen(1))
+			Expect(string(*(*result.Instances)[0].DeletionStatus)).To(Equal("PENDING"))
```
gabriel-farache, line 552: I missed that in the previous PR but `PENDING` state may be misleading. For me pending means "in queue for creation" and I am pretty sure the SPs are setting "PENDING" status after their POST request completed

The name of the const is `DeletionStatusPending` why not rename it `PENDING_DELETION" or "DELETION_PENDING"? (in a follow up PR)

machacekondra (approved):

internal/service/resource_manager/service_type_instance.go
```diff
@@ -186,7 +186,25 @@ func (s *InstanceService) DeleteInstance(ctx context.Context, instanceID string,
...
+				log.Error("Failed to reset retry count for instance", "instance_id", instanceID, "error", resetErr)
+			}
+		} else {
+			// Mark as pending deletion
+			if markErr := s.store.ServiceTypeInstance().MarkForDeletion(ctx, instanceID); markErr != nil {
+				return service.NewInternalError(fmt.Sprintf("failed to mark instance %s for deletion: %v", instanceID, markErr))
+			}
+		}
+
+		log.Info("Deferred deletion of instance from provider", "instance_id", instance.ID, "provider_name", instance.ProviderName)
```
gciavarrini: If `deferred=true` we skip the provider delete call; the log still sounds like deletion happened at the provider, can you please reword it?

gciavarrini (commented):
Do we need to update the enhancement as well?

internal/service/resource_manager/service_type_instance_test.go
ygalblum: As explained in https://github.com/dcm-project/cli/pull/20#discussion_r3064488511, these are two separate fields. I don't see a problem with two separate fields using the same value as the field defines the meaning of the status. Also, currently, the deletion status may also be `Failed` when the retry count is exceeded.

internal/service/resource_manager/service_type_instance_test.go
ygalblum: Three lines down: https://github.com/dcm-project/service-provider-manager/pull/51/changes#diff-731937135a71f6612d4f4726ccda587fc5072543bf780139dadccb4daa2be536R519

internal/service/resource_manager/service_type_instance_test.go
gabriel-farache: ok for this version 
We may want to change that to have a list of conditions as you mentioned K8s is doing
And would `DEFERRED` be a more consistent word rather than `PENDING`?

gabriel-farache (approved):

internal/service/resource_manager/service_type_instance.go
ygalblum: Add `Scheduled`: 8b1d62a

ygalblum:
> Do we need to update the enhancement as well?

https://github.com/dcm-project/enhancements/pull/42

gciavarrini (approved):

PR #52, opened 2026-04-09
Makefile - Find the installed compose tool instead of hard coding it

ygalblum:


jenniferubah (approved):

gabriel-farache (approved):
<3

machacekondra (approved):

gciavarrini (approved):
We need to update other Makefiles as well. I'll create a story for it

PR #56, opened 2026-04-17
Persist provider status field in the data model

ygalblum:
## What
Add the `status` field (registered/updated) to the provider data model so it is persisted and returned in all API responses.

## Why
The OpenAPI spec defines a read-only `status` field on the Provider schema, but the value was only computed transiently during registration and never stored. As a result, GET, List, and PUT responses omitted the field entirely.

## Changes
- Added `Status` field to the provider database model with a default of `registered`
- Updated model-to-API conversion to read status from the persisted model
- Set status to `registered` on creation and `updated` on any modification
- Added `Status` to the explicit column select list in the store update method
- Removed the now-unnecessary `ModelToProviderWithStatus` helper

## Testing
Added status assertions to existing service and handler tests for Get, List, and Update operations, verifying both `registered` and `updated` status values are correctly persisted and returned.

### DIFF
--- a/internal/handlers/provider/handler_test.go
+++ b/internal/handlers/provider/handler_test.go
@@ -139,7 +139,7 @@ var _ = Describe("Handler", func() {
 			Expect(*jsonResp.Providers).To(BeEmpty())
 		})
 
-		It("returns providers", func() {
+		It("returns providers with status", func() {
 			// Create providers first
 			for _, name := range []string{"provider-1", "provider-2"} {
 				createReq := providerserver.CreateProviderRequestObject{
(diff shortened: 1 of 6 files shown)


gabriel-farache:
Do we really need to store the status? It makes no sense to me to store a status "UPDATED" without the diff; instead of persisting it we could just hard code the status in the handler/service when the endpoints are called and complete successfully (for POST and PUT); this was done for `RegisterOrUpdateProvider` (POST) but missing from GET/LIST/PUT if I am not wrong

If we were to keep in DB the deleted SP then it would make sense to have the status persisted, otherwise I do not see any value

ygalblum:
> hard code the status in the handler/service when the endpoints are called and complete

Yes, it's missing in GET and LIST. If it's not persistent in the DB then how would you handle the following sequence:

1. Create SP A
2. Update SP A
3. Restart the Service Provider Manager
4. Get SP A is called - How will the code know that update was called on SP A?

Now, I'm not arguing that it must be in the object schema. We might as well remove it from there. But, as long as it's there, it needs to be persisted in the DB

machacekondra:
We talked about this in one of the first PRs. @gciavarrini decided to leave it as is now, for some reason which I don't remember. But yes, the point is that the status is useless. To me it's better remove then return something useless.

ygalblum:
Makes sense. Closing this one and opened another PR for the removal: #57

PR #58, opened 2026-04-21
Add PENDING_PROVIDER deletion status for provider-aware deferred cleanup

ygalblum:
Add a PENDING_PROVIDER deletion status that parks deferred deletions when their provider is unhealthy, preventing wasted retries against unavailable providers.

Previously, when a provider went down, the cleanup scheduler would burn through retries against it, eventually marking instances as FAILED. When the provider recovered, those instances stayed FAILED with no automatic recovery path.

- Added PENDING_PROVIDER to the deletion_status enum in the OpenAPI spec and store constants
- Added bulk store methods to transition instances between PENDING/FAILED and PENDING_PROVIDER by provider name
- Added an atomic store method using a provider health subquery to park individual instances, preventing race conditions between the scheduler and monitor
- Extended the healthcheck monitor to transition deletion statuses on provider health changes (Ready→NotReady parks, NotReady→Ready reactivates)
- Extended the cleanup scheduler to check provider health before each deletion attempt as a self-correcting safety net
- Modernized existing monitor code to use Go builtins (min/max) and tagged switch

- Added store-layer tests for the three new methods including cross-provider isolation and idempotency.
- Added monitor tests for deletion status transitions on health changes.
- Added scheduler tests for provider health checking before deletion.
- Added an E2E test for the full PENDING_PROVIDER lifecycle.


api/v1alpha1/resource_manager/openapi.yaml
```diff
@@ -293,7 +293,9 @@ components:
             Deletion status for deferred deletions. Absent for active
             instances. PENDING indicates the instance is queued for cleanup.
             FAILED indicates the cleanup has exceeded maximum retries.
-          enum: [PENDING, FAILED]
+            PENDING_PROVIDER indicates the instance is waiting for its
+            provider to become healthy before cleanup is retried.
+          enum: [PENDING, FAILED, PENDING_PROVIDER]
```
jenniferubah: If `PENDING` is used when the provider is unreachable and `PENDING_PROVIDER` is used when the provider is unhealthy, could we use one (preference in `PENDING_PROVIDER`) for both cases? Since both cases are dependent on the Provider hence it is a bit confusing, just the state of the provider is different (IIUC).

ygalblum: I guess I wasn't clear to begin with. So, please help me make it more readable. The values for deletion status are:

- `PENDING` - scheduled for deletion, regardless of the state of the provider. The cleanup thread looks for the instances with this deletion status to call the provider's delete endpoint. 
- `FAILED` - the cleanup thread keeps a retry count. The instance's deletion status is set to FAILED once the retry is exhausted. 
- `PENDING_PROVIDER` - The SPRM knows that the SP is not available and changes the deletion status accordingly. This allows the cleanup to skip trying to delete it during this period. In addition, once the SP is back on, the deletion status is turned back to PENDING for the cleanup to pick it up again.

jenniferubah: Got it now, thanks. I misunderstood.  So maybe we can the enum status [`SCHEDULED, FAILED` , `PENDING_PROVIDER`]. This way the status explictly tells what it it. WDYT?

ygalblum: Done: a091e8f

internal/cleanup/scheduler.go
```diff
@@ -85,6 +85,17 @@ func (s *Scheduler) ProcessPendingDeletions(ctx context.Context) {
...
 func (s *Scheduler) processOne(ctx context.Context, instance model.ServiceTypeInstance) {
 	log := logging.FromContext(ctx)
+
+	parked, err := s.store.ServiceTypeInstance().MarkPendingProviderIfNotReady(ctx, instance.ID)
+	if err != nil {
+		log.Error("Failed to check provider health for instance", "instance_id", instance.ID, "error", err)
+		return
+	}
+	if parked {
+		log.Info("Instance parked as PENDING_PROVIDER, provider is not ready", "instance_id", instance.ID, "provider_name", instance.ProviderName)
```
jenniferubah: nit: "Instance `parked` or `marked`...."? Also not sure if the variable was intentionally named `parked` `?

internal/healthcheck/monitor.go
```diff
@@ -121,6 +124,17 @@ func (m *Monitor) checkProvider(ctx context.Context, provider model.Provider) {
 			"old_status", provider.HealthStatus,
 			"new_status", newStatus,
 		)
+
+		switch newStatus {
+		case model.HealthStatusNotReady:
+			if err := m.instanceStore.MarkProviderDeletionsPendingProvider(ctx, provider.Name); err != nil {
+				slog.Error("Failed to park deletions for unhealthy provider", "provider", provider.Name, "error", err)
```
jenniferubah, line 131: Same as above

jenniferubah (approved):
Small nit about wordings `park` and `mark`

gciavarrini (approved):
I agree with Jennifer comments and please also verify if we need to update the enhancement too. 
Overall the PR LGTM

internal/cleanup/scheduler.go
ygalblum: Actually, I did mean `parked`. Because the deletion process is suspended and the instance is parked in this state. But, if you think `marked` is better (or any other word) I don't mind changing it

jenniferubah: I prefer `marked`. It reads better in my head.

ygalblum: Done: e674625

jenniferubah: Thanks!

jenniferubah:
@ygalblum please don't forget to backport this to the release branch

ygalblum:
https://github.com/dcm-project/service-provider-manager/pull/59

PR #61, opened 2026-04-27
Add `service_type` Filter to Resource Manager List Endpoint

ebichman-1:
## What

Added a service_type query parameter to GET /service-type-instances so users can filter instances by service type (e.g., vm, container). 

## Why

Users currently can only filter instances by provider name. To find all instances of a given service type (e.g., all VMs), they
must know which providers serve that type and issue separate queries per provider. 
A service_type filter lets them query directly, 
For example: 
```bash
GET /service-type-instances?service_type=vm
```
without needing provider-level knowledge.

## Changes

- Add ```service_type``` query parameter to the openapi spec for the list endpoint.
- Add indexed ```ServiceType``` column to the ```ServiceTypeInstance``` DB model (non-null, defaults to '').
- Add ```ServiceType``` to ```ServiceTypeInstanceListOptions``` and apply a filter clause in the store layer.
- Update ```ListInstances``` to accept the new parameter, extract and persist ```service_type``` from spec during ```CreateInstance```.
- Pass ```ServiceType``` from the request params through the handler to the service.
- Validate that ```spec.service_type``` is present, is a string, and is non-empty on instance creation.

#### Added tests across all layers:

- Store: ```filter-by-service-type``` test using ```newServiceTypeInstanceWithType``` helper.
- Service: ```filter-by-service-type test```, ```CreateInstance``` sets ```service_type``` from spec, validation error tests for missing, ```non-string```, empty, and ```whitespace-only``` ```spec.service_type```.
- Handler: ```filter-by-service-type``` test with multiple ```providers/types```, validation error tests for missing, ```non-string```, and empty ```spec.service_type```.
- E2E: all existing instance creation specs updated to include ```service_type``` in spec.


api/v1alpha1/resource_manager/openapi.yaml
```diff
@@ -281,6 +286,11 @@ components:
           type: string
           description: Name of the provider
           example: "kubevirt-123"
+        service_type:
```
ygalblum: While we want to be able to filter based on `service_type` we don't want to add it to the schema. The SPRM should assume the information is stored in the `spec`, even though it is not strictly defined in the latter's schema.

We have a similar issue (of assuming the undefined internals) in the Policy Manager. While the `spec` is not strictly defined: https://github.com/dcm-project/policy-manager/blob/c10123346e0a31ed69ff80d9afdf67f910be7129/api/v1alpha1/engine/openapi.yaml#L62, the code does look into the payload: https://github.com/dcm-project/policy-manager/blob/c10123346e0a31ed69ff80d9afdf67f910be7129/internal/handlers/engine/converters.go#L33.

So, what this service should do is get this information from the spec and store it in the DB. However, it should should use it only for filtering and not return it as part of the payload.

Furthermore, the DB field should be set at `index` to allow for quicker filtering. A similar approach is taken in this service for the `ServiceType` of the `CatalogItem`: https://github.com/dcm-project/catalog-manager/blob/1efc0c43a886131f6a7c6d7e9cc5f01f594dd2a5/internal/store/model/catalog_item.go#L19 and the `CatalogItem` of the `CatalogItemInstance`: https://github.com/dcm-project/catalog-manager/blob/1efc0c43a886131f6a7c6d7e9cc5f01f594dd2a5/internal/store/model/catalog_item_instance.go#L19. In both cases, the information is stored twice (as a field and inside the `spec`) but returned only once. The difference here is that it is part of the typed schema.

We can have a later discussion on whether common `spec` fields (https://github.com/dcm-project/catalog-manager/blob/1efc0c43a886131f6a7c6d7e9cc5f01f594dd2a5/api/v1alpha1/servicetypes/common.yaml#L85) should be known to the internal services

ebichman-1: Sure, I understand your point.

Let’s discuss the details tomorrow when you’re available. 
Then, I'll clarify the task’s goal in the Jira description to ensure that the solution is aligned with what you’ve mentioned in the latest comment.

internal/config/config.go
```diff
@@ -37,6 +37,8 @@ type DBConfig struct {
 	Name     string `envconfig:"DB_NAME" default:"service-provider"`
 	User     string `envconfig:"DB_USER"`
 	Password string `envconfig:"DB_PASS"`
+	// SSLMode is passed to Postgres as sslmode (libpq). Use require or verify-full in production.
```
ygalblum: Why is this part of this commit?

ygalblum (changes requested):
The configuration flag does not belong to this PR. Otherwise LGTM

ygalblum:
Also, when squashing and merging a PR, the commit message it taken from the PR description . So, it shouldn't contain all this information. Please rewrite it to summarize the change

internal/config/config.go
ebichman-1: wow, my bad, sorry for that. I'll remove it.

ygalblum (approved):

internal/handlers/resource_manager/handler_test.go
```diff
@@ -230,6 +230,56 @@ var _ = Describe("Resource Manager Handler", func() {
 			Expect(*jsonResp.Instances).To(HaveLen(3))
 		})
 
+		It("filters by service_type", func() {
```
gciavarrini, line 281: Can you add a test case for a missing/wrong-typed `service_type`? And another test for non string `service_type` ?

internal/service/resource_manager/service_type_instance.go
```diff
@@ -79,10 +79,17 @@ func (s *InstanceService) CreateInstance(ctx context.Context, request *resource_
 		return nil, service.NewProviderError(fmt.Sprintf("Error from Provider (%s): %v", providerName, err))
 	}
 
+	// Extract service_type from spec for indexed filtering
+	var serviceType string
+	if st, ok := request.Spec["service_type"].(string); ok {
```
gciavarrini: What if `ok == false`?

gciavarrini (changes requested):
Some changes are needed. 

Please ping me once the technical issues are resolved for a second look.

In addition, the git history and commit messages must be cleaned up before this PR can be considered for merging.
For example, the current commit messages contain redundant 'Signed-off-by' tags, typos, and outdated implementation details that do not reflect the current code.

**After** you have the approvals, you must clean the git history. 
I want to perform a final review of the clean state before merging this PR.

internal/service/resource_manager/service_type_instance.go
ebichman-1: This scenario currently triggers a validation Error (status_code: 400).
we have two ways to handle this: 

1. Maintain the current behavior: we can return the 400 error directly to the user.
2. Adjust validation: we can validate only strings values and allow to insert empty string into db. 

 What would you recommend as the best approach here?

gciavarrini: I don't agree with 
> This scenario currently triggers a validation Error (status_code: 400).

if `ok==false` the `serviceType` var remains empty without any error, so we store a `service_type == ""`.
`ok ==false` if `request.Spec["service_type"]` is absent or if it's present but not a string (e.g. boolean, object, array).

internal/service/resource_manager/service_type_instance.go
```diff
@@ -72,19 +72,19 @@ func (s *InstanceService) CreateInstance(ctx context.Context, request *resource_
 		return nil, err
 	}
 
+	// Extract service_type from spec — ok==false if the key is absent or the value is not a string
+	serviceType, ok := request.Spec["service_type"].(string)
+	if !ok {
+		return nil, service.NewValidationError("spec.service_type is required and must be a string")
+	}
```
gciavarrini, line 80: @ygalblum I'd like to have your opinion for this. 
Do you think it's ok to enforce this check? Also if service_type is empty the validation pass.

internal/service/resource_manager/service_type_instance_test.go
gciavarrini, line 1: missing tests for wrong/missing service_type

gciavarrini (changes requested):

internal/service/resource_manager/service_type_instance.go
ygalblum: Yes, I think we should enforce it.

 As I wrote here: https://github.com/dcm-project/service-provider-manager/pull/61#discussion_r3156555462
```
We can have a later discussion on whether common spec fields (https://github.com/dcm-project/catalog-manager/blob/1efc0c43a886131f6a7c6d7e9cc5f01f594dd2a5/api/v1alpha1/servicetypes/common.yaml#L85) should be known to the internal services
```
If we has added it to begin with, we would have gotten the enforcement already. But, as I wrote, I don't want to do it as a side effect of this PR.

So yes, we should also check that it's not an empty string.

internal/service/resource_manager/service_type_instance_test.go
ebichman-1: Thanks to you and @ygalblum for the feedback, I’ll definitely include those tests for the service_type parameter.

ebichman-1: I've just pushed a new commit that includes the missing validation check for an empty string, and its corresponding tests.

internal/handlers/resource_manager/handler_test.go
```diff
@@ -158,6 +158,54 @@ var _ = Describe("Resource Manager Handler", func() {
...
+				},
+			}
+
+			resp, err := handler.CreateInstance(ctx, req)
+
+			Expect(err).NotTo(HaveOccurred())
+			_, ok := resp.(server.CreateInstance400ApplicationProblemPlusJSONResponse)
+			Expect(ok).To(BeTrue())
+			Expect(providerCalled).To(BeFalse())
+		})
```
gciavarrini, line 208: Please add a test to verify the behavior for `service_type= " "` (whitespace only)

gciavarrini (changes requested):
Also still not addressed @ygalblum https://github.com/dcm-project/service-provider-manager/pull/61#issuecomment-4353731772

internal/handlers/resource_manager/handler_test.go
ebichman-1: Sure, agreed, I'll add this test too

ebichman-1: Sure, I'll squash and merge all the latest commits too, 
Thank you for all of your feedback comments

gciavarrini: I don't see the test. Can you please link it if i'm missing it

ebichman-1: lease take a look at this line 
```go
It("returns validation error when spec.service_type is whitespace only...
```

gciavarrini: I was looking in `handler_test.go`, it's in [service_type_instance_test.go](https://github.com/ebichman-1/dcm-service-provider-manager/blob/d8165d46839325a33c291f1fd5673fb4f6636468/internal/service/resource_manager/service_type_instance_test.go#L328-L343). That’s sufficient for the behavior we care about, so I’m okay without a separate handler test.

gciavarrini (approved):
Approved **ONLY** from a technical standpoint. 

**Hold merge until**:
1 - DCO check passes
2 - fix commit typos (Assitend) and use consistent commit attribution (e.g. Assistend-By vs Co-authored-by), 
3 - short and clear PR description for squash-merge (as per @ygalblum [comment](https://github.com/dcm-project/service-provider-manager/pull/61#issuecomment-4353731772))

**I want to review these point before merging**.

ebichman-1:
@gciavarrini Sure, I’ll perform another squash and commit with a message that complies with the DCO script. 
- I’ll also ensure the commit follows the required standards 
- I will streamline the PR description to be more concise.

.github/workflows/check-aep.yaml
```diff
@@ -2,7 +2,7 @@ name: Check AEP Compliance
 
 on:
   pull_request:
-    branches: [main, 'release/v*']
+    branches: [main]
```
ygalblum: Something went wrong in the rebase here (and in other workflows)

ygalblum (changes requested):
Need to revert the change to the workflow files. But, otherwise LGTM

.github/workflows/check-aep.yaml
ebichman-1: @ygalblum, I see it now, Looks like a side effect of the rebase when squashing the commits into one.
I'll try to remove these changes manually.

ygalblum (approved):

gciavarrini (approved):
lgtm

PR #65, opened 2026-05-01
Add three-state health model for service providers

ygalblum:
## What
Replace the binary Ready/NotReady health model with a three-state model (Ready, Unhealthy, Unavailable) to distinguish between an unreachable service provider and one whose backing provider is unavailable.

## Why
The previous two-state model could not differentiate between a service provider that is down and one that is up but whose backing provider is unavailable. This distinction is needed so DCM can make better scheduling decisions.

## Changes
- Replace HealthStatusNotReady with HealthStatusUnhealthy (reachable but backing provider unavailable) and HealthStatusUnavailable (unreachable)
- Parse the health endpoint JSON response body to determine backing provider health via the status field (healthy/unhealthy)
- Transition to Unhealthy immediately on 200 OK with status "unhealthy", without requiring consecutive failures
- Apply exponential backoff only for Unavailable providers; Unhealthy providers use the standard check interval since they are reachable
- Park deletions on transition to either Unhealthy or Unavailable, reactivate on recovery to Ready
- Update store query to match both Unhealthy and Unavailable statuses when marking pending provider deletions

## Testing
Added unit tests for the Unhealthy state across monitor, store, and cleanup scheduler packages, covering health check transitions, backoff scheduling, deletion parking, and reactivation.

### DIFF
--- a/internal/healthcheck/monitor.go
+++ b/internal/healthcheck/monitor.go
@@ -4,7 +4,6 @@ package healthcheck
 import (
 	"context"
 	"encoding/json"
-	"io"
 	"log/slog"
 	"math"
 	"net/http"
@@ -178,14 +177,8 @@ func (m *Monitor) performHealthCheck(ctx context.Context, provider model.Provide
 		return healthCheckFailed
 	}
 
-	body, err := io.ReadAll(resp.Body)
-	if err != nil {
-		slog.Error("Error reading health check response body", "provider", provider.Name, "error", err)
-		return healthCheckFailed
-	}
-
 	var hr healthResponse
-	if err := json.Unmarshal(body, &hr); err != nil {
+	if err := json.NewDecoder(resp.Body).Decode(&hr); err != nil {
 		slog.Error("Error parsing health check response", "provider", provider.Name, "error", err)
 		return healthCheckFailed
 	}
--- a/internal/healthcheck/monitor.go
+++ b/internal/healthcheck/monitor.go
@@ -3,6 +3,8 @@ package healthcheck
 
 import (
 	"context"
+	"encoding/json"
+	"io"
 	"log/slog"
 	"math"
 	"net/http"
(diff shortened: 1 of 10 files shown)


ygalblum:
Depends on the approval of https://github.com/dcm-project/enhancements/pull/47

ygalblum:
The Enhancement PR was approved. But, this one should go in only after all SPs report the health of the their backend:
* Kubevirt: https://github.com/dcm-project/kubevirt-service-provider/pull/28
* K8S Container: https://github.com/dcm-project/k8s-container-service-provider/pull/28
* 3 Layer App: https://github.com/dcm-project/three-tier-app-demo-service-provider/pull/22
* Cluster - already reports `unhealth` when the backing cluster is not ready

ygalblum:
Now that all SPs report the health of their backend this PR is ready

gciavarrini (approved):
lgtm

internal/healthcheck/monitor.go
```diff
@@ -138,36 +158,56 @@ func (m *Monitor) checkProvider(ctx context.Context, provider model.Provider) {
...
+	if err != nil {
+		slog.Error("Error reading health check response body", "provider", provider.Name, "error", err)
+		return healthCheckFailed
+	}
+
+	var hr healthResponse
+	if err := json.Unmarshal(body, &hr); err != nil {
+		slog.Error("Error parsing health check response", "provider", provider.Name, "error", err)
+		return healthCheckFailed
+	}
```
jenniferubah: nit: instead of io, we could used json decoder for the response body, no?
`json.NewDecoder(resp.Body).Decode(&hr); `

jenniferubah (approved):
Overall, this is good. One minor comment

internal/healthcheck/monitor.go
ygalblum: Good catch. Fixed

PR #68, opened 2026-06-03
Reset health backoff when provider re-registers

chadcrum:
## Summary
- Clear `ConsecutiveFailures` and `NextHealthCheck` when a provider re-registers via `RegisterOrUpdateProvider`, so the health monitor picks it up on the next tick instead of leaving it stuck behind a stale backoff window
- Scoped to re-registration only — the `UpdateProvider` (PATCH/PUT) path preserves health state so routine field updates don't silently reset backoff
- `HealthStatus` is intentionally preserved; the health monitor confirms health before promoting back to "ready"

## Test plan
- [x] Unit test: re-register resets `ConsecutiveFailures` to 0 and `NextHealthCheck` to nil
- [x] Unit test: `UpdateProvider` preserves existing health backoff state
- [x] Full suite passes (`go test ./...`)
- [x] Deploy to test cluster with `quay.io/chadcrum0/test:service-provider-manager-FLPATH-4331_sprm-reset-health-on-reregister` and verify provider recovery after restart

Ref: [FLPATH-4331](https://issues.redhat.com/browse/FLPATH-4331)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/service/provider/provider.go
+++ b/internal/service/provider/provider.go
@@ -55,6 +55,8 @@ func (s *ProviderService) RegisterOrUpdateProvider(ctx context.Context, req *pro
 
 	if existing != nil {
 		log.Debug("Existing provider found, updating", "provider_id", existing.ID, "name", req.Name)
+		existing.ConsecutiveFailures = 0
+		existing.NextHealthCheck = nil
 		updated, err := s.updateExistingProvider(ctx, existing, req)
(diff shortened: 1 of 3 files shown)


gabriel-farache (approved):

gciavarrini (approved):
lgtm

